PR #33, opened 2021-09-08
Update org name for kube-vip project

abhay-krishna:
Kube-vip updated their org from plunder-app to kube-vip, so we should do it for consistency.
/hold


TerryHowe (approved):
/lgtm

PR #39, opened 2021-09-09
Fix build arg for base image

abhay-krishna:
Use new variable that was defined for cluster-controller base image.


pokearu (approved):
/lgtm

PR #43, opened 2021-09-14
work around codebuild gopath shennanigans by pre-installing controlle…

danbudris:
…r gen, rather than allowing troubleshoot makefile to do it and set the gopath incorrectly

So, we call `make support-bundle` in order to build the support bundle binaries.

In the [troubleshoot `Makefile` target `support-bundle`](https://github.com/replicatedhq/troubleshoot/blob/master/Makefile), they check for the presence of `controller-gen` and `client-gen` using `which`; if they don't exist in the path, they'll install them and then call them based on the gopath. However, because of the way codebuild structures the gopath for sub-modules, this isn't going to work correctly. So, we instead pre-install them and ensure that the top-level codebuild gopath/bin dir is on the path before calling `make support-bundle`.

*Issue #, if available:*

*Description of changes:*


jaxesn (approved):
/lgtm

PR #46, opened 2021-09-15
update base image comment to reflect reality

danbudris:
update base image comment to reflect reality


vivek-koppuru (approved):
/lgtm

jaxesn (commented):
Can we also add the yum clean in the same layer as installing tar? Minor thing but good practice. Check one of our other images for an example

PR #197, opened 2021-11-23
move adding users/groups to container image to common

jaxesn:
*Issue #, if available:*

*Description of changes:*
There are a few container images where we add new users.  This standardizes that to a common target in Common.mk.  This should also help with reproducibility of container images using the buildkitd inline cache option.


abhay-krishna (approved):
/lgtm

PR #226, opened 2021-12-02
Break helm build and push into two scripts

TerryHowe:
Attempting to simplify previous PR

### DIFF
--- a/projects/aws/eks-anywhere-test/Makefile
+++ b/projects/aws/eks-anywhere-test/Makefile
@@ -16,8 +16,12 @@ RELEASE_TARGETS=images
 
 include $(BASE_DIRECTORY)/Common.mk
 
+copy-attribution:
+	@cp $(ATTRIBUTION_TARGETS) $(OUTPUT_DIR)
+	@echo Done
+
 .PHONY: local-images
-local-images: eks-anywhere-test/images/amd64 helm/build
+local-images: copy-attribution eks-anywhere-test/images/amd64 helm/build
 	
 .PHONY: images
-images: eks-anywhere-test/images/push helm/push
+images: copy-attribution eks-anywhere-test/images/push helm/push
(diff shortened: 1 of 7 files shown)


projects/aws/eks-anywhere-test/Makefile
```diff
@@ -17,8 +16,12 @@ RELEASE_TARGETS=images
 
 include $(BASE_DIRECTORY)/Common.mk
 
+copy-attribution:
```
jaxesn, line 19: Why the empty attribution? If has_licenses was true then this additional copy Target wouldn't be necessary since it automatically does this before building images. That wouldn't work tho since there is no go code for it to try and gather licenses from.

TerryHowe: I couldn't get attributions to work. Maybe I should remove this and create a ticket to work on that later

jaxesn: What wouldn't work? Generating one or was it complaining somewhere that it didn't exist? It can't really generate one since it depends on using go-licenses to gather licenses from a go source tree which this project doesn't have.

TerryHowe: I made attributes optional for now and this works. I created an issue to fix attributions for this project. I really want these changes for *other* projects.

Out of the box the attribution generation fails trying to checkout the GIT_TAG of the non-existent repo

jaxesn: Right, attribution generation is dependent on a golang source project, typically from cloning upstream.  I think I see what you were trying to do now with the attribution file in the helm script.  I have a thought Ill try out and see how it looks.

jaxesn (approved):
This looks good to me. the tink stuff will keep failing until that is merged. Feel free to manually merge this.

/lgtm

PR #232, opened 2021-12-03
eks-anywhere-test: cleanup attribution handling for helm chart

jaxesn:
*Issue #, if available:*

*Description of changes:*
Follow up to #226, an attempt to improve the Attribution file handling.  Since we want to include the attribution in the helm chart, we can copy it into the output dir before calling helm just like we do for images.


TerryHowe (approved):
/lgtm

PR #237, opened 2021-12-04
Cert manager helm chart

TerryHowe:
The helm chart is copied from the source repo, but slightly modified so we can set our registry instead of quay.

```
tlhowe@3c22fb41240e eks-anywhere-test % helm list
NAME                        	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART                                                                 	APP VERSION                                    
cert-manager-1639449732     	default  	1       	2021-12-13 19:42:17.977306 -0700 MST	deployed	cert-manager-v1.1.0-f3025dd61112d09578b35bfe41ab845131baa0ae-helm     	v1.1.0-f3025dd61112d09578b35bfe41ab845131baa0ae
eks-anywhere-test-1639449795	default  	1       	2021-12-13 19:43:22.023971 -0700 MST	deployed	eks-anywhere-test-v0.1.1-f3025dd61112d09578b35bfe41ab845131baa0ae-helm	v0.1.1-f3025dd61112d09578b35bfe41ab845131baa0ae
tlhowe@3c22fb41240e eks-anywhere-test % k get pods
NAME                                                  READY   STATUS    RESTARTS   AGE
cert-manager-1639449732-cainjector-7568dc8d96-wxm4k   1/1     Running   0          89s
cert-manager-1639449732-df68f496c-tl69h               1/1     Running   0          89s
cert-manager-1639449732-webhook-78d47f755-9dt5c       1/1     Running   0          89s
eks-anywhere-test-65f7cfbfc5-d2cth                    1/1     Running   0          25s
```


Common.mk
```diff
@@ -135,6 +135,10 @@ BINARY_TARGETS?=$(call BINARY_TARGETS_FROM_FILES_PLATFORMS, $(BINARY_TARGET_FILE
 BINARY_TARGET_FILES?=
 SOURCE_PATTERNS?=.
 
+#################### HELM ######################
```
jaxesn, line 138: You should be able to just set HAS_HELM_CHART in the cert manager makefile instead of introducing these new vars.

TerryHowe: Not sure how I missed that. Done.

jaxesn:
its very possible you did this and I just missed it, but for these helm charts that we are copying from upstream, maintenance will be become a nightmare. Could we put in a place a process for clearly marking the aws changes in the code or in the readme (if we really dont want to clutter the code) so that when we need to update its a simpler diff process?

Common.mk
```diff
@@ -411,15 +438,16 @@ validate-checksums: $(BINARY_TARGETS)
 
 # Build helm chart
 .PHONY: helm/build
-
+helm/build: ## Build helm chart
+helm/build: $(if $(or $(wildcard $(PROJECT_ROOT)/helm_patches),$(wildcard $(MAKE_ROOT)/helm_patches)),$(HELM_GIT_PATCH_TARGET),$(HELM_GIT_CHECKOUT_TARGET))
```
jaxesn, line 442: Move the attribution prereq to this line.

Common.mk
```diff
@@ -307,6 +317,23 @@ $(GIT_PATCH_TARGET): $(GIT_CHECKOUT_TARGET)
 	$(BASE_DIRECTORY)/build/lib/go_mod_download.sh $(MAKE_ROOT) $(REPO) $(GIT_TAG) $(GOLANG_VERSION) $(REPO_SUBPATH)
 	@touch $@
 
+ifneq ($(HAS_HELM_CHART),true)
```
jaxesn, line 320:
```suggestion
ifeq ($(HAS_HELM_CHART),true)
```

Common.mk
```diff
@@ -307,6 +317,23 @@ $(GIT_PATCH_TARGET): $(GIT_CHECKOUT_TARGET)
...
+	@echo rm -f $(HELM_REPOSITORY)/eks-anywhere-*
+	(cd $(HELM_REPOSITORY) && $(BASE_DIRECTORY)/build/lib/wait_for_tag.sh $(HELM_GIT_TAG))
+	git -C $(HELM_REPOSITORY) checkout -f $(HELM_GIT_TAG)
+	touch $@
+endif
+
+$(HELM_GIT_PATCH_TARGET): $(HELM_GIT_CHECKOUT_TARGET)
+	git -C $(HELM_REPOSITORY) config user.email prow@amazonaws.com
+	git -C $(HELM_REPOSITORY) config user.name "Prow Bot"
+	git -C $(HELM_REPOSITORY) am --committer-date-is-author-date $(or $(wildcard $(PROJECT_ROOT)/helm_patches),$(wildcard $(MAKE_ROOT)/patches))/*
```
jaxesn, line 334:
```suggestion
	git -C $(HELM_REPOSITORY) am --committer-date-is-author-date $(or $(wildcard $(PROJECT_ROOT)/helm_patches),$(wildcard $(MAKE_ROOT)/helm_patches))/*
```

projects/jetstack/cert-manager/helm/Chart.yaml.template
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
```
jaxesn, line 1: Will these removed now that you are cloning?

jaxesn (changes requested):
I like this!

Common.mk
TerryHowe: Done.

Common.mk
TerryHowe: Moved to helm/patches

Common.mk
TerryHowe: I removed it from the helm/push target if that is what was causing the confusion.

projects/jetstack/cert-manager/helm/Chart.yaml.template
TerryHowe: Yeh, I am going to remove these templates and replace them with a search and replace thing.

TerryHowe (commented):
Pushing new stuff soon

Common.mk
```diff
@@ -112,6 +112,7 @@ BUILDER_IMAGE?=$(BASE_IMAGE_REPO)/$(BASE_IMAGE_NAME)-builder:$(BASE_IMAGE_TAG)
 ####################################################
 
 #################### IMAGES ########################
+HAS_HELM_CHART?=false
```
jaxesn, line 115: How about just moving this down to the HELM block?

Common.mk
```diff
@@ -307,6 +316,23 @@ $(GIT_PATCH_TARGET): $(GIT_CHECKOUT_TARGET)
...
+$(HELM_GIT_CHECKOUT_TARGET): | $(HELM_REPOSITORY)
+	@echo rm -f $(HELM_REPOSITORY)/eks-anywhere-*
+	(cd $(HELM_REPOSITORY) && $(BASE_DIRECTORY)/build/lib/wait_for_tag.sh $(HELM_GIT_TAG))
+	git -C $(HELM_REPOSITORY) checkout -f $(HELM_GIT_TAG)
+	touch $@
+
+$(HELM_GIT_PATCH_TARGET): $(HELM_GIT_CHECKOUT_TARGET)
+	git -C $(HELM_REPOSITORY) config user.email prow@amazonaws.com
+	git -C $(HELM_REPOSITORY) config user.name "Prow Bot"
+	git -C $(HELM_REPOSITORY) am --committer-date-is-author-date $(or $(wildcard $(PROJECT_ROOT)/helm/patches),$(wildcard $(MAKE_ROOT)/patches))/*
```
jaxesn, line 332: I think you want to change the last patches to helm/patches as well. These two wildcards are to cover patches folders in the root and in release branch folders, which ever may exists if either.

Common.mk
```diff
@@ -410,17 +436,20 @@ validate-checksums: $(BINARY_TARGETS)
 	$(BUILDCTL)
 
 # Build helm chart
+ifeq ($(HAS_HELM_CHART),true)
 .PHONY: helm/build
-
+helm/build: ## Build helm chart
+helm/build: $(if $(or $(wildcard $(PROJECT_ROOT)/helm/patches),$(wildcard $(MAKE_ROOT)/helm/patches)),$(HELM_GIT_PATCH_TARGET),$(HELM_GIT_CHECKOUT_TARGET))
```
jaxesn, line 442: I was suggestion just moving `$(OUTPUT_DIR)/ATTRIBUTION.txt` to this line before the if instead of two pre-req lines.  Its a nit, we just usually put all the pre-reqs on the same line, unless they are variable sets.

Common.mk
```diff
@@ -135,12 +136,20 @@ IMAGE_USERADD_USER_NAME?=
...
 LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) 
 IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push)) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) 
 ####################################################
 
+#################### HELM ##########################
+HELM_SOURCE_REPOSITORY?=$(REPO_OWNER)/$(REPO)
+HELM_GIT_TAG?=$(GIT_TAG)
+HELM_DIRECTORY?=.
+HELM_REPOSITORY?=$(REPO)
+HELM_GIT_CHECKOUT_TARGET?=$(REPO)/eks-anywhere-checkout-$(HELM_GIT_TAG)
```
jaxesn, line 149: I think you want to replace REPO with HELM_REPOSITORY in these two vars?

build/lib/helm_build.sh
```diff
@@ -18,18 +18,37 @@ set -o errexit
...
+# Copy
+#
+mkdir -p ${DEST_DIR}
+cp ${OUTPUT_DIR}/ATTRIBUTION.txt ${DEST_DIR}/
+cp -r ${SOURCE_DIR} ${DEST_DIR}
+
+#
+# Search and replace
+#
+SEDFILE=${OUTPUT_DIR}/helm/sedfile
```
jaxesn, line 39: Are you breaking this out to that helm_template script below?

projects/jetstack/cert-manager/README.md
```diff
@@ -17,3 +17,9 @@ You can find the latest versions of these images on ECR Public Gallery.
 [cert-manager Controller](https://gallery.ecr.aws/eks-anywhere/jetstack/cert-manager-controller) | 
 [CA injector](https://gallery.ecr.aws/eks-anywhere/jetstack/cert-manager-cainjector) | 
 [cert-manager Webhook Server](https://gallery.ecr.aws/eks-anywhere/jetstack/cert-manager-webhook)
+
+### Helm Chart
+
+The helm chart is a modified version of the source helm chart located in the jetstck/cert-manager repo at `deploy/charts/cert-manager/`.
+
+If there are any patches to the make file, use `git format-patch $(cat ../GIT_TAG)` and add them to the `patches` directory.
```
jaxesn, line 25: `helm/patches`?

projects/jetstack/cert-manager/helm/sedfile.template
```diff
@@ -0,0 +1,3 @@
+s,quay.io,${HELM_REGISTRY},g
```
jaxesn, line 1: I guess an alternative would be to copy the upstream values.yaml and make whatever changes we want there, instead of dynamically making the change with sed at build time?  Helm doesnt support multiple values.yaml does it where we could apply our specific values on top of the upstream?

I dont have a strong opinion against what you are doing, just trying to wrap my head around it a bit.

jaxesn (changes requested):
mostly nits, love the progress!

projects/jetstack/cert-manager/helm/sedfile.template
TerryHowe: The sed search and replace is mainly to replace registry and image tag at build time because those depend on the actual build, but I tried to leave it flexible in case we want to use that for something else.

Common.mk
TerryHowe: This gets used in the targets before the helm chart. I guess I could put the helm block before that. It was that way for a bit.

Common.mk
TerryHowe: Yeh, good call. I'd removed support for PROJECT_ROOT patches from the other place, so I'm going to pull it here. MAKE_ROOT helm/patches only at least for now

Common.mk
TerryHowe: I'm not terribly attached, but as long as it acts the same, I think it is a bit easier to read on three lines.

Common.mk
TerryHowe: Yeh, good call

TerryHowe:
build kit can't open some files, looks like some disk issue. Retesting

build/lib/helm_build.sh
TerryHowe: The other script is gone now

projects/jetstack/cert-manager/helm/patches/0001-Minor-clean-up-to-helm-chart.patch
```diff
@@ -0,0 +1,25 @@
+From 3263a4c1fbf6fb2398665008ed010b0c1b3c6950 Mon Sep 17 00:00:00 2001
+From: Terry Howe <tlhowe@amazon.com>
```
TerryHowe, line 2: upstream https://github.com/jetstack/cert-manager/pull/4656

Common.mk
jaxesn: It shouldn't matter where it is in the file vs other places its used since just about everything in here is ?= so its delayed evaluation anyway.

Common.mk
```diff
@@ -410,17 +438,21 @@ validate-checksums: $(BINARY_TARGETS)
 	$(BUILDCTL)
 
 # Build helm chart
+ifeq ($(HAS_HELM_CHART),true)
 .PHONY: helm/build
-
+helm/build: ## Build helm chart
 helm/build: $(OUTPUT_DIR)/ATTRIBUTION.txt
```
jaxesn, line 444: Does this work separating the pre-reqs like this?  if so, yeah it is def cleaner.

Common.mk
```diff
@@ -410,17 +438,21 @@ validate-checksums: $(BINARY_TARGETS)
...
-	$(BUILD_LIB)/helm_build.sh $(IMAGE_REPO) $(IMAGE_COMPONENT) $(IMAGE_TAG) $(OUTPUT_DIR)
+helm/build: $(if $(REPO_NO_CLONE),,$(HELM_GIT_CHECKOUT_TARGET))
```
jaxesn, line 445: I know its ugly, but i think to be most correct we need to make sure this equals true, with the whole filter statement we do everywhere else... i think there is def room for improvement there, but probably best to stick with it until we change it more widely?

Common.mk
```diff
@@ -410,17 +438,21 @@ validate-checksums: $(BINARY_TARGETS)
...
+helm/build: $(if $(wildcard $(MAKE_ROOT)/helm/patches),$(HELM_GIT_PATCH_TARGET))
```
jaxesn, line 446: Usually I would write it like this `$(if $(wildcard $(MAKE_ROOT)/helm/patches),$(HELM_GIT_PATCH_TARGET),)` with the trailing comma to indicate if the if evals to false nothing is returned.  I guess make is smart enough to assume that?

projects/jetstack/cert-manager/helm/sedfile.template
jaxesn: Yeah that makes sense, I was just wondering if helm supported multiple values files to avoid having to do this at all and let helm take care of it?

projects/aws/eks-anywhere-test/Makefile
```diff
@@ -15,8 +14,13 @@ REPO_NO_CLONE=true
 BUILD_TARGETS=local-images
 RELEASE_TARGETS=images
 
+HAS_HELM_CHART=true
+
 include $(BASE_DIRECTORY)/Common.mk
 
+.PHONY: clean-repo
```
jaxesn, line 21: Instead of this how about we change the preq in Common.mk to something like

```
.PHONY: clean
clean: $(if has_no_repo....,clean-repo,)
```

jaxesn (changes requested):

projects/aws/eks-anywhere-test/Makefile
TerryHowe: Done

projects/jetstack/cert-manager/helm/sedfile.template
TerryHowe: My understanding is there are 3 values files in order of what has say:
* Users values file overrides everything
* Parent chart values file which overrides the default
* This values file which is the default

We could also always override the values as the "user", but I think for security purposes, we want it so when we install v1.1.0-1 cert manager it installs the v1.1.0-1 container images all the time. We could in theory make every add-on a nested chart, but that seems messy for what we want to do.  We won't allow the user to override the images in a helm chart using the controller, but obviously, the user can do whatever they like if they were using the chart outside of that.

jaxesn: ok yeah that makes sense. Thanks for the explanation.

jaxesn (approved):
Looks really good!

/lgtm

PR #244, opened 2021-12-07
Remove trailing parenthesis in script

abhay-krishna:
This was causing the crictl URL to be suffixed with a parenthesis at the end.


jaxesn (approved):
Ah damn, I did the same thing on kind but the presubmit caught it

/lgtm

PR #280, opened 2021-12-15
kind: add haproxy image now that capd allows overwriting

jaxesn:
*Issue #, if available:*

*Description of changes:*
This introduces a haproxy image which can be used with capd instead of pulling from gcr/docker.  This also adds the new kind base image which will be handled like our other base images for updates and what not.


abhay-krishna (approved):
/lgtm

PR #283, opened 2021-12-16
helm: new project build for helm cli

jaxesn:
*Issue #, if available:*

*Description of changes:*

We *may* want this in the cli tools image for cilium upgrades.  Still need the prow jobs if we want to go this route.

@g-gaston will decide if we want this vs using the go library instead.


TerryHowe (approved):
/lgtm
/hold 
The hold is just in case someone else has comments

PR #286, opened 2021-12-21
licenses: no need to keep code

jaxesn:
*Issue #, if available:*

*Description of changes:*
go-licenses is pretty conservative about copying code for projects with certain licenses since they "could" have been modified.  We know this is not the case since we are calling go mod vendor to pull from upstream before building.


TerryHowe (approved):
/lgtm

PR #298, opened 2021-12-23
capi: remove coredns migration patch

jaxesn:
*Issue #, if available:*

*Description of changes:*
In capi 0.3.18 they updated this dependency to a released version of 1.0.12 which had not be released at the time when we first started using capi.  We could/should have removed this patch a while again when we bumped to 0.3.19 back in June, but it was missed.  This hasnt been an issues because we control the coredns version based on eks-d, but just in last weeks release we bumped to 1.8.4 which has triggered this issue.

This will need to be backported to the release-0.6 branch

Note: I did not rename all the other patches because 0001 was deleted.  There is no issue with starting at 0002 other than just how it looks in the folder.  Next time we significantly change the patches we can rename them.


vivek-koppuru (approved):
/lgtm

PR #300, opened 2021-12-23
fix help generation on builder base

jaxesn:
*Issue #, if available:*

*Description of changes:*
Should fix yesterdays failure due to a difference in bash on builder-base and mac.  This also

- separates checksum + attribution generation
- moves create_pr to be after each part (help, attribution, help) the script should still work for this and this would allow for anything that passes to still get checked in vs if one fails they all fail.

### DIFF
--- a/build/lib/generate_help_body.sh
+++ b/build/lib/generate_help_body.sh
@@ -80,7 +80,7 @@ if [[ "$HAS_PATCHES" == "true" ]]; then
 fi
 
 IMAGE_TARGETS_HELP=""
-if [ ! -z "$LOCAL_IMAGE_TARGETS" ]; then
+if [ ! -z "$(echo "$LOCAL_IMAGE_TARGETS" | xargs)" ]; then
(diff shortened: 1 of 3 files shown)


build/lib/generate_help_body.sh
```diff
@@ -80,7 +80,7 @@ if [[ "$HAS_PATCHES" == "true" ]]; then
 fi
 
 IMAGE_TARGETS_HELP=""
-if [ ! -z "$LOCAL_IMAGE_TARGETS" ]; then
+if [ ! -z "$(echo "$LOCAL_IMAGE_TARGETS" | xargs)" ]; then
```
TerryHowe, line 83: not sure what xargs would do here

TerryHowe (approved):
/lgtm

build/lib/generate_help_body.sh
jaxesn: it trims the string of trailing spaces.. its a hack

PR #305, opened 2021-12-23
handle nonstandard attribution file names and ignored files

jaxesn:
*Issue #, if available:*

*Description of changes:*

We havent been updating KINDNETD_ATTRIBUTION.txt when it needed. This also avoids the error when trying to added ignored files by using the check-ignore trick used for the Makefile git add.


TerryHowe (approved):
/lgtm

PR #308, opened 2021-12-27
Use whatever repo we are supposed to and docker config.json

TerryHowe:
Change the helm build so

* Push helm chart to IMAGE_REPO (build doesn't have perms to get pub repo)
* Do not overlay the value of DOCKER_CONFIG
* Clean up the clean up


jaxesn (approved):
/lgtm

PR #312, opened 2021-12-29
Add helm to license gather

TerryHowe:
Gather licenses for helm repositories if they are not the original source repository


Common.mk
```diff
@@ -153,6 +153,7 @@ HELM_DESTINATION_REPOSITORY?=$(IMAGE_COMPONENT)
 HELM_ADDITIONAL_KEY_VALUES?=
 HELM_GIT_CHECKOUT_TARGET?=$(HELM_SOURCE_REPOSITORY)/eks-anywhere-checkout-$(HELM_GIT_TAG)
 HELM_GIT_PATCH_TARGET?=$(HELM_SOURCE_REPOSITORY)/eks-anywhere-helm-patched
+HELM_LICENSES_TARGET=$(OUTPUT_DIR)/LICENSES/github.com/$(HELM_SOURCE_OWNER)/$(HELM_SOURCE_REPOSITORY)/COPYING
```
jaxesn, line 156: This seems pretty specific, I assume the flux/cert-manager have a COPYING file?  The license gathering code will look for a various versions of LICENSE/COPYRIGHT/COPYING/etc...  I'm fine to go with this for now, worst case it a project doesn't have this file this target will just run every-time gather-licenses is called, which isn't that big of a deal.

TerryHowe: Hmm, this is copy/pasta. Maybe I didn't choose a good example.

jaxesn (approved):
/lgtm

PR #313, opened 2021-12-30
Chart licenses

TerryHowe:
- Gathering licenses for helm charts and putting them in containers didn't make much sense, so I've reverted that.
- Gathered licenses just for the helm chart and put them in the helm chart. (works for cert-manager and flux)
- Removed attribute file from helm chart since that is really for the container image

### DIFF
--- a/build/lib/common.sh
+++ b/build/lib/common.sh
@@ -179,10 +179,9 @@ function build::non-golang::gather_licenses(){
 function build::non-golang::copy_licenses(){
   local -r source_dir="$1"
   local -r destination_dir="$2"
-  cd $source_dir
-  license_files=($(find . \( -name "*COPYING*" -o -name "*COPYRIGHT*" -o -name "*LICEN[C|S]E*" -o -name "*NOTICE*" \)))
-  cd -
-  for file in "${license_files[@]}"; do
+  (cd $source_dir; find . \( -name "*COPYING*" -o -name "*COPYRIGHT*" -o -name "*LICEN[C|S]E*" -o -name "*NOTICE*" \)) |
+  while read file
+  do
     license_dest=$destination_dir/$(dirname $file)
     mkdir -p $license_dest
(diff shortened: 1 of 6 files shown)


build/lib/common.sh
```diff
@@ -181,6 +181,19 @@ function build::non-golang::gather_licenses(){
   rm -rf $project_name
 }
 
+function build::non-golang::copy_licenses(){
```
jaxesn, line 184: Can we standardize this for the above method somehow?  Haven't looked too closely at the changes but could this be made so it could be called by gather_licenses after it clones/checksout the repo?

build/lib/helm_build.sh
```diff
@@ -27,12 +27,15 @@ CHART_NAME=$(basename ${HELM_DESTINATION_REPOSITORY})
...
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+source "${SCRIPT_ROOT}/common.sh"
+
 #
 # Copy
 #
 mkdir -p ${DEST_DIR}
-cp ${OUTPUT_DIR}/ATTRIBUTION.txt ${DEST_DIR}/
 cp -r ${SOURCE_DIR} ${DEST_DIR}
+build::non-golang::copy_licenses ${HELM_SOURCE_REPOSITORY} $DEST_DIR/LICENSES/github.com/${HELM_SOURCE_REPOSITORY}
```
jaxesn, line 38: So this get bundled in the chart?

build/lib/common.sh
TerryHowe: Done

build/lib/helm_build.sh
TerryHowe: Yes, the chart contains the licenses from the helm repo

jaxesn (approved):
/lgtm

PR #315, opened 2022-01-04
Add patch to retain update permission on etcdadmcluster for KCP controller

mrajashree:
When we added external etcd support, we had modified the KCP controller's role.yaml to grant it permissions on the etcdadmcluster object. One of the permissions was `update`, for the upgrade workflow, which was added in by [this patch](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kubernetes-sigs/cluster-api/patches/0017-Synchronize-upgrade-flow.patch#L103-L120). This got missed while rebasing all commits on the new [1.0.1 branch](https://github.com/mrajashree/cluster-api). This PR adds back this permission required for the upgrade use case

*Issue #, if available:*

*Description of changes:*


vivek-koppuru (approved):
/lgtm

PR #316, opened 2022-01-04
Build Redis image

TerryHowe:
Redis build. I still need to add the repository.

/hold


Common.mk
```diff
@@ -156,6 +156,7 @@ HELM_GIT_PATCH_TARGET?=$(HELM_SOURCE_REPOSITORY)/eks-anywhere-helm-patched
 ####################################################
 
 #################### BINARIES ######################
+GOLANG_VERSION?=""
```
jaxesn, line 159: shouldnt need the quotes and you can remove them in the if below as well.

build/lib/common.sh
```diff
@@ -188,6 +188,30 @@ function build::non-golang::copy_licenses(){
   done
 }
 
+function build::non-golang::generate_attribution(){
```
jaxesn, line 191: Im not sure this is going to fly honestly.  Worth a shot for sure, but we'll need to get this generated and then create the proper oss ticket to review it and make sure it is complete enough.

jaxesn: Looking at the resulting file it does seem reasonable tho.  If we get this approved for redis, we should circle back and add this to the source-controller build which gathers nongolang licenses as well and get them added into our attribution.

projects/redis/redis/ATTRIBUTION.txt
```diff
@@ -0,0 +1,298 @@
...
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------
+
+** github/redis/redis; version 6.2.6 --
```
jaxesn, line 85: This could be an issue since lua probably has its own git repo and version...

projects/redis/redis/Makefile
```diff
@@ -0,0 +1,25 @@
...
+REPO_OWNER=redis
+IMAGE_COMPONENT=redis/redis
+IMAGE_BUILD_ARGS?=IMAGE_TAG
+
+BASE_IMAGE_NAME?=eks-distro-minimal-base-glibc
+SIMPLE_CREATE_BINARIES=false
+GO_MOD_DOWNLOAD_TARGETS=$(GIT_PATCH_TARGET)
+HAS_GO_LICENSES=false
+
+BUILD_TARGETS=local-images
```
jaxesn, line 14: Can we remove these? Is a target running that wont work without these overrides?

jaxesn (changes requested):
This looks pretty good!  I am mostly just concerned on the attribution stuff since this is really our first time with a non-golang build so we just haven't been thru the oss office paces yet.  We should start that now to make sure we don't need to adjust course.

projects/redis/redis/docker/linux/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+ARG BASE_IMAGE
+ARG BUILDER_IMAGE
+FROM $BUILDER_IMAGE as redis-builder
+
+COPY redis /redis
+RUN set -x && yum install -y gcc make
+WORKDIR /redis
+RUN case "$(uname -p)" in \
```
jaxesn, line 22: Another option to this would be just to use the redis provided by al2. I suspect its really out of date so may not be the best option, but maybe when al2022 is out we could rely on the redis provided there and just stay aligned with that?  That would sorta push off some of the maintenance from us since we could rely on them to keep it up to date and backport cves and what not. Just a thought.

build/lib/common.sh
TerryHowe: The format at least is similar to what `generate-attribution` creates.

Redis is odd in that it has snapshots of dependencies in the repo, so this works as it is (kind of). We really need a list of repos and git tags to gather licenses. My only idea is we manually create the file and use it as input along with this script.

Common.mk
TerryHowe: Done.

projects/redis/redis/Makefile
TerryHowe: Done.

projects/redis/redis/docker/linux/Dockerfile
TerryHowe: Darn, I didn't know they had it and v6.2.6 which is the latest. I'll switch to that.

TerryHowe: Switched to AL2 redis

jaxesn (changes requested):
This looks great!  I think we still need a presubmit here?  I'd like to see the list of OS packages installed for redis, which I can get from a presubmit log once it exists.

I think this one is kind of an odd ball in the way we are using it from al2 vs building it, which i think is the right approach, but it means some of the stuff like the GIT_TAG and ATTRIBUTION.txt file are kind of unnecessary.  It doesnt bother me, just throwing it out there.

TerryHowe:
Yeh, there is probably some package clean-up that could happen

https://github.com/aws/eks-anywhere-prow-jobs/pull/71

TerryHowe:
Removed check-sums, binaries and attribution targets from build

projects/redis/redis/docker/linux/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ARG BUILDER_IMAGE
+FROM $BUILDER_IMAGE as redis-builder
+
+WORKDIR /
+RUN set -x && \
+    amazon-linux-extras enable redis6 && \
+    cp /etc/yum.repos.d/amzn2-extras.repo /newroot/etc/yum.repos.d/amzn2-extras.repo && \
+    clean_install systemd true true && \
+    clean_install redis && \
+    remove_package systemd true && \
```
jaxesn, line 25: I wonder if we can clean this up some?  Typically in cases like this in the base images, we end up removing a number of packages that are not needed during runtime, such as bash/gawk/sed/other utils/etc.  I wonder if we know redis doesn't need those at run time?  I usually ldd on the binary to check for libraries it is loaded, but *mostly* you can remove the same packages we remove for nginx, something like [this](https://github.com/aws/eks-distro-build-tooling/blob/main/eks-distro-base/Dockerfile.minimal-base-nginx#L37) + bash (most likely)

For reference here are the packages currently included from the build log

```
#7 40.80 + cat /tmp/packages/redis
#7 40.80 audit-libs-2.8.1-3.amzn2.1.x86_64
#7 40.80 basesystem-10.0-7.amzn2.0.1.noarch
#7 40.80 bash-4.2.46-34.amzn2.x86_64
#7 40.80 bzip2-libs-1.0.6-13.amzn2.0.3.x86_64
#7 40.80 ca-certificates-2021.2.50-72.amzn2.0.1.noarch
#7 40.80 coreutils-8.22-24.amzn2.x86_64
#7 40.80 cracklib-2.9.0-11.amzn2.0.2.x86_64
#7 40.80 cracklib-dicts-2.9.0-11.amzn2.0.2.x86_64
#7 40.80 elfutils-default-yama-scope-0.176-2.amzn2.noarch
#7 40.80 elfutils-libelf-0.176-2.amzn2.x86_64
#7 40.80 elfutils-libs-0.176-2.amzn2.x86_64
#7 40.80 filesystem-3.2-25.amzn2.0.4.x86_64
#7 40.80 gawk-4.0.2-4.amzn2.1.2.x86_64
#7 40.80 glibc-2.26-56.amzn2.x86_64
#7 40.80 glibc-common-2.26-56.amzn2.x86_64
#7 40.80 glibc-minimal-langpack-2.26-56.amzn2.x86_64
#7 40.80 gmp-6.0.0-15.amzn2.0.2.x86_64
#7 40.80 gpg-pubkey-c87f5b1a-593863f8
#7 40.80 grep-2.20-3.amzn2.0.2.x86_64
#7 40.80 gzip-1.5-10.amzn2.x86_64
#7 40.80 info-5.1-5.amzn2.x86_64
#7 40.80 keyutils-libs-1.5.8-3.amzn2.0.2.x86_64
#7 40.80 krb5-libs-1.15.1-37.amzn2.2.2.x86_64
#7 40.80 libacl-2.2.51-14.amzn2.x86_64
#7 40.80 libattr-2.4.46-12.amzn2.0.2.x86_64
#7 40.80 libcap-2.22-9.amzn2.0.2.x86_64
#7 40.80 libcap-ng-0.7.5-4.amzn2.0.4.x86_64
#7 40.80 libcom_err-1.42.9-19.amzn2.x86_64
#7 40.80 libcrypt-2.26-56.amzn2.x86_64
#7 40.80 libdb-5.3.21-24.amzn2.0.3.x86_64
#7 40.80 libgcc-7.3.1-13.amzn2.x86_64
#7 40.80 libgcrypt-1.5.3-14.amzn2.0.2.x86_64
#7 40.80 libgpg-error-1.12-3.amzn2.0.3.x86_64
#7 40.80 libpwquality-1.2.3-5.amzn2.x86_64
#7 40.80 libselinux-2.5-12.amzn2.0.2.x86_64
#7 40.80 libsemanage-2.5-11.amzn2.x86_64
#7 40.80 libsepol-2.5-8.1.amzn2.0.2.x86_64
#7 40.80 libstdc++-7.3.1-13.amzn2.x86_64
#7 40.80 libverto-0.2.5-4.amzn2.0.2.x86_64
#7 40.80 logrotate-3.8.6-15.amzn2.x86_64
#7 40.80 lz4-1.7.5-2.amzn2.0.1.x86_64
#7 40.80 ncurses-6.0-8.20170212.amzn2.1.3.x86_64
#7 40.80 ncurses-base-6.0-8.20170212.amzn2.1.3.noarch
#7 40.80 ncurses-libs-6.0-8.20170212.amzn2.1.3.x86_64
#7 40.80 openssl-libs-1.0.2k-19.amzn2.0.10.x86_64
#7 40.80 pam-1.1.8-23.amzn2.0.1.x86_64
#7 40.80 pcre-8.32-17.amzn2.0.2.x86_64
#7 40.80 popt-1.13-16.amzn2.0.2.x86_64
#7 40.80 redis-6.2.6-1.amzn2.x86_64
#7 40.80 sed-4.2.2-5.amzn2.0.2.x86_64
#7 40.80 setup-2.8.71-10.amzn2.0.1.noarch
#7 40.80 shadow-utils-4.1.5.1-24.amzn2.0.2.x86_64
#7 40.80 system-release-2-14.amzn2.x86_64
#7 40.80 systemd-libs-219-78.amzn2.0.15.x86_64
#7 40.80 tzdata-2021a-1.amzn2.noarch
#7 40.80 ustr-1.0.4-16.amzn2.0.3.x86_64
#7 40.80 xz-libs-5.2.2-1.amzn2.0.2.x86_64
#7 40.80 zlib-1.2.7-18.amzn2.x86_64
```

bash + sed + shadow-utils + ncurses + gawk + grep stick out as possibly not needed.  What I have found before is most of these deps get included either from bash, or because they are needed during package install time, like shadow-utils, but not at runtime.

projects/redis/redis/docker/linux/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN set -x && \
+    sed -i -e 's/^logfile .*/logfile ""/' /newroot/etc/redis/redis.conf
+
+FROM $BASE_IMAGE
+ARG IMAGE_TAG=not-set
+ENV REDIS_VERSION $IMAGE_TAG
+
+COPY --from=redis-builder /newroot /
+
+WORKDIR /
```
jaxesn, line 36: Im assuming when redis is installed it will create a user, you check the /etc/passwd|group files to see.  We may want to set that user as a docker User command?

Youll have to double check, I *think* since we are installing these packages using --installroot it should update the /newroot/etc/passwd|group file, but make sure and if its modifying the /etc/ files instead, wed need to make those changes to the /newroot/etc ones.

jaxesn (changes requested):
This looks really good.  I put some thoughts just around being consistent with how we handle other images like this made up of AL2 packages, wrt to cleaning up unnecessary packages.  This is an effort to limit our scope of installed packages and then therefore the kinds of CVEs we can expect from OS packages.

projects/redis/redis/docker/linux/Dockerfile
TerryHowe: I still need to investigate this user situation

TerryHowe: Should be good to go now

jaxesn (approved):
/lgtm

PR #319, opened 2022-01-06
Wait until tarball is available under latest before downloading

abhay-krishna:
This function will make the job sleep for 20 mins or until the project tarball is available under `latest/`, whichever is earlier.


vivek-koppuru (approved):
/lgtm

PR #328, opened 2022-01-11
Set bash as shell for make in common makefile

g-gaston:
*Description of changes:*
By default make's shell is sh, which doesn't support flags like -o pipefail


abhay-krishna (approved):
/lgtm

TerryHowe (approved):

PR #329, opened 2022-01-11
Bump cilium to v1.9.11 and add cilium chart digest

g-gaston:
*Description of changes:*
To download the oci manifest I'm using [this](https://github.com/csweichel/oci-tool). I have to do a bit of research and see if that's the most canonical tool. In theory, any cli that is able to download the raw oci manifest would work for us.

I tried with docker, but it doesn't respect the media types (it overrides all of them to docker types). This changes the manifest and hence the checksum.


jaxesn (commented):
Can you add to the readme about oci-tool and installing it for updating?  Weve been adding updating sections to the readmes of projects weve updated recently.  Check out the flux one for a reference, something like would be cool...

jaxesn (approved):
Im good, thanks for the readme!

abhay-krishna (approved):
/lgtm

PR #336, opened 2022-01-13
Add status.version to list of fields to ignore for update

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Corresponding PR to mrajashree/cluster-api repo: https://github.com/mrajashree/cluster-api/pull/9

When `KubeadmControlPlane` was upgraded to `v1beta1`, the `status.version` field was added which was called out here: https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/api/v1alpha3/conversion.go#L98

Because of this, with the `Update` function we are taking advantage of in our controller, the validation webhook for `KubeadmControlPlane` is failing saying that changes to this field is forbidden. By adding it to the list of paths to ignore, we can get past this error which is specific to this field.

When making the commit to upstream, either we go with this approach or edit the conversion logic somehow to be able to understand how to get the value of the version from the spec potentially when converting from `v1alpha3` to `v1beta1`. For now though, we can go with this approach and figure out what upstream maintainers prefer as the approach.


vivek-koppuru:
/hold

Testing

TerryHowe (approved):
/lgtm

PR #344, opened 2022-01-20
Set namespace arg to eksa-system for capi, capv and external etcd con…

mrajashree:
…trollers

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/projects/kubernetes-sigs/cluster-api-provider-vsphere/build/create_manifests.sh
+++ b/projects/kubernetes-sigs/cluster-api-provider-vsphere/build/create_manifests.sh
@@ -30,8 +30,7 @@ source "${MAKE_ROOT}/../../../build/lib/common.sh"
 
 cd $REPO
 
-sed -i '' 's/^\( *\) - --enable-leader-election$/&\
-\1 - --namespace=eksa-system/' ./config/manager/manager.yaml
+yq eval -i -P ".spec.template.spec.containers[0].args += [\"--namespace=eksa-system\"]" config/manager/manager.yaml
 
 make manifests STAGE="release" \
   MANIFEST_DIR="out" \
--- a/projects/kubernetes-sigs/cluster-api-provider-vsphere/build/create_manifests.sh
+++ b/projects/kubernetes-sigs/cluster-api-provider-vsphere/build/create_manifests.sh
@@ -30,6 +30,9 @@ source "${MAKE_ROOT}/../../../build/lib/common.sh"
 
 cd $REPO
 
+sed -i '' 's/^\( *\) - --enable-leader-election$/&\
+\1 - --namespace=eksa-system/' ./config/manager/manager.yaml
+
 make manifests STAGE="release" \
   MANIFEST_DIR="out" \
   PULL_POLICY="IfNotPresent" \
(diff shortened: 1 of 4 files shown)


projects/kubernetes-sigs/cluster-api/build/create_manifests.sh
```diff
@@ -50,14 +50,26 @@ make set-manifest-pull-policy PULL_POLICY=IfNotPresent TARGET_RESOURCE="./config
...
+
 ## Build the manifests
 make release-manifests
 
 ## Build the development manifests
 make -C test/infrastructure/docker set-manifest-image \
 MANIFEST_IMG=$CAPI_REGISTRY_PREFIX/capd-manager MANIFEST_TAG=$IMAGE_TAG
 make -C test/infrastructure/docker set-manifest-pull-policy PULL_POLICY=IfNotPresent
-PATH="$(pwd)/hack/tools/bin:$PATH" make -C test/infrastructure/docker release-manifests 
+sed -i '' 's/^\( *\) - "--leader-elect"$/&\
```
abhay-krishna, line 69: `sed -i ''`
This syntax of `sed` inplace will work only on MacOS. Using `gsed` instead would also us to use the same syntax as Nix systems.

```suggestion
SED=sed
if [ "$(uname -s)" = "Darwin" ]; then
    SED=gsed
fi
$SED -i '' 's/^\( *\) - "--leader-elect"$/&\
```

projects/kubernetes-sigs/cluster-api-provider-vsphere/build/create_manifests.sh
```diff
@@ -30,6 +30,9 @@ source "${MAKE_ROOT}/../../../build/lib/common.sh"
 
 cd $REPO
 
+sed -i '' 's/^\( *\) - --enable-leader-election$/&\
```
abhay-krishna, line 33: `sed -i ''`
This syntax of `sed` inplace will work only on MacOS. Using `gsed` instead would allow us to use the same syntax as Nix systems.

```suggestion
SED=sed
if [ "$(uname -s)" = "Darwin" ]; then
    SED=gsed
fi
$SED -i '' 's/^\( *\) - "--leader-elect"$/&\
```

Same below

abhay-krishna:
This looks fine and works, but I would suggest using `yq` as it's a more straightforward and cleaner approach and it's clear what we're trying to do.
For example,
```bash
yq eval -i -P ".spec.template.spec.containers[0].args += [\"--namespace=eksa-system\"]" config/manager/manager.yaml
```
Although we could argue that not everyone running this script locally may have `yq` installed, installing it may be a better option than getting into the weeds of `sed`.

jaxesn:
Yeah I agree with @abhay-krishna  lets use yq for this.  When running locally, a number of projects in this repo require yq to load the eks-d manifest so I think its safe to introduce it here.

mrajashree:
> For example,
> 
> yq eval -i -P ".spec.template.spec.containers[0].args += [\"--namespace=eksa-system\"]" config/manager/manager.yaml

yeah this looks much cleaner and there's no concerns of the upstream manifest changing. will switch to yq

abhay-krishna (approved):
/lgtm

mrajashree:
https://github.com/aws/eks-anywhere/issues/810

PR #346, opened 2022-01-20
patch troubleshoot to avoid nil pointer when namespace does not exist

danbudris:
*Issue #, if available:*

*Description of changes:*

Since upgrading troubleshoot, it looks like when a deployment analyzer is run targeting a namespace which does not exist, the returned analyze result is nil.
The converter does not handle nil values properly and will attempt to process them regardless, causing a nil pointer exception.
Addressing this via a patch to ensure our support bundle functionality is working properly while working on an upstream fix, which will hopefully get into the next release and we remove this patch and bump the ts version.


danbudris:
will update the checksums once available in the build

projects/replicatedhq/troubleshoot/patches/0002-Add-nil-check-to-analyzer-processing.patch
```diff
@@ -0,0 +1,98 @@
...
+-			Meta: Meta{
+-				Name: name,
+-				Labels: map[string]string{
+-					"desiredPosition": "1",
+-					"iconKey":         i.IconKey,
+-					"iconUri":         i.IconURI,
+-				},
+-			},
+-			Insight: &Insight{
++		if i != nil {
```
TerryHowe, line 34: Wouldn't a `if i == nil { continue }` be a bit smaller a change? Really this seems to be fine though

danbudris: yeah, it would be smaller, tho all of this is basically just an indent

danbudris: simplified it, will need to update checksums once presubmits run

TerryHowe (approved):
/lgtm

TerryHowe:
Yeh, checksums

jaxesn (approved):
/lgtm

PR #349, opened 2022-01-20
upload tink artifacts to the s3 bucket

panktishah26:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/projects/tinkerbell/tink/expected_artifacts
+++ b/projects/tinkerbell/tink/expected_artifacts
@@ -1,0 +1,13 @@
+SHA256SUM
+SHA256SUM.sha256
+SHA256SUM.sha512
+SHA512SUM
+SHA512SUM.sha256
+SHA512SUM.sha512
+githash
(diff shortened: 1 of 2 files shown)


projects/tinkerbell/tink/expected_artifacts
```diff
@@ -0,0 +1,13 @@
+SHA256SUM
+SHA256SUM.sha256
+SHA256SUM.sha512
+SHA512SUM
+SHA512SUM.sha256
+SHA512SUM.sha512
+githash
+tink-linux-amd64-b89f887c4f7d1c47809548ef0e3446dab8129c88.tar.gz
```
abhay-krishna, line 8: You can use $GIT_TAG here instead of the actual tag or commit, it will be substituted when validating the artifacts. Refer to how we've done it in the other projects, for example, CAPI.
```suggestion
tink-linux-amd64-$GIT_TAG.tar.gz
```

PR #359, opened 2022-01-22
Add etcdadm cipher suites flag support for bottlerocket

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Adding support for taking in cipher suites flag to pass to bottlerocket-bootstrap container. 

Created an issue to fix how we are passing in the flags to the etcdadm run command: https://github.com/aws/eks-anywhere/issues/1020

This PR should also be held until the tags are updated for etcdadm merged from this: https://github.com/mrajashree/etcdadm-bootstrap-provider/pull/9


mrajashree (approved):
/lgtm

danbudris (approved):

abhay-krishna:
Not a blocker but it would be nice to have the READMEs and the UPSTREAM_PROJECTS.yaml updated with the new tags.

vivek-koppuru:
> Not a blocker but it would be nice to have the READMEs and the UPSTREAM_PROJECTS.yaml updated with the new tags.

Also the attibrution file, can follow up with PR

PR #368, opened 2022-01-26
chartmuseum

vincentni:
*Issue #, if available:*
`go-licenses` failed to find some licenses

> one or more libraries have an incompatible/unknown license: map["unknown":["github.com/NetEase-Object-Storage/nos-golang-sdk/config" "github.com/NetEase-Object-Storage/nos-golang-sdk/logger" "github.com/NetEase-Object-Storage/nos-golang-sdk/noserror" "github.com/NetEase-Object-Storage/nos-golang-sdk/utils" "github.com/NetEase-Object-Storage/nos-golang-sdk/model" "github.com/NetEase-Object-Storage/nos-golang-sdk/nosconst" "github.com/NetEase-Object-Storage/nos-golang-sdk/nosclient" "github.com/NetEase-Object-Storage/nos-golang-sdk/auth" "github.com/NetEase-Object-Storage/nos-golang-sdk/httpclient"]]

*Description of changes:*
chartmuseum binary


Common.mk
```diff
@@ -318,7 +318,7 @@ $(GIT_PATCH_TARGET): $(GIT_CHECKOUT_TARGET)
 	git -C $(REPO) am --committer-date-is-author-date $(PATCHES_DIR)/*
 	@touch $@
 
-%eks-anywhere-go-mod-download: $(if $(PATCHES_DIR),$(GIT_PATCH_TARGET),$(GIT_CHECKOUT_TARGET))
+%eks-anywhere-go-mod-download: $(if $(PATCHES_DIR),$(GIT_PATCH_TARGET),$(GIT_CHECKOUT_TARGET)) $(PRECOMPILE_PREPARE)
```
jaxesn, line 321: Id prefer not doing it this way, ill add a note below on a different options.

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
+BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
+GIT_TAG?=$(shell cat GIT_TAG)
+GOLANG_VERSION?="1.17"
+
+REPO=harbor
+REPO_OWNER=goharbor
+REPO_SUBPATH=./src
+
+BASE_IMAGE_NAME?=eks-distro-base
```
jaxesn, line 9: I think probably because of that entrypoint we are using the standard base here which has bash, but Id really prefer not to... Anyway we could avoid the entrypoint?  Or maybe we need to make a minimal image with bash that we use only in specific places like this.

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
+BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
+GIT_TAG?=$(shell cat GIT_TAG)
+GOLANG_VERSION?="1.17"
+
+REPO=harbor
+REPO_OWNER=goharbor
+REPO_SUBPATH=./src
```
jaxesn, line 7: Im not sure why this is set?

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+BASE_IMAGE_NAME?=eks-distro-base
+DOCKERFILE_FOLDER=./docker/linux/$(IMAGE_NAME)
```
jaxesn, line 10: I believe this can go

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+FETCH_BINARIES_TARGETS=eksa/helm/chartmuseum
+
+IMAGE_NAMES=harbor-chartserver
+
+HARBOR_CHARTSERVER_IMAGE_COMPONENT=harbor/harbor-chartserver
```
jaxesn, line 16: I think here just set IMAGE_COMPONENT

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+$(call IMAGE_TARGETS_FOR_NAME, harbor-chartserver): harbor-chartserver-useradd/images/export
+
+harbor-chartserver-useradd/images/export: IMAGE_USERADD_USER_NAME=chart IMAGE_USERADD_USER_ID=10000
+
+PRECOMPILE_PREPARE=swagger_prepare
+
+DOCKERFILEPATH=$(PROJECT_ROOT)/$(REPO)/make/photon
```
jaxesn, line 24: Is this used?

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+DOCKERFILEPATH_CHART_SERVER=$(DOCKERFILEPATH)/chartserver
+
+harbor_base_namespace=goharbor
```
jaxesn, line 27: Lets put this in all caps jsut to match pattern

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+BASE_IMAGE_NAME?=eks-distro-base
+DOCKERFILE_FOLDER=./docker/linux/$(IMAGE_NAME)
+
+FETCH_BINARIES_TARGETS=eksa/helm/chartmuseum
+
+IMAGE_NAMES=harbor-chartserver
+
+HARBOR_CHARTSERVER_IMAGE_COMPONENT=harbor/harbor-chartserver
+
+$(call IMAGE_TARGETS_FOR_NAME, harbor-chartserver): harbor-chartserver-useradd/images/export
```
jaxesn, line 18: Move these to after the include of common.mk

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+harbor-chartserver-useradd/images/export: IMAGE_USERADD_USER_NAME=chart IMAGE_USERADD_USER_ID=10000
+
+PRECOMPILE_PREPARE=swagger_prepare
```
jaxesn, line 22: Instead of this, after doing the include lets try this

```
$(call IMAGE_TARGETS_FOR_NAME, harbor-chartserver): swagger_prepare

```

If possible, Id prefer swagger_prepare be a real file path if there is one that is getting generated when its called.

projects/goharbor/harbor/docker/linux/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ARG TARGETARCH
+ARG TARGETOS
+
+RUN yum install -y shadow-utils >> /dev/null \
+    && yum clean all \
+    && groupadd -f -r -g 10000 chart \
+    && useradd --no-log-init -m -g 10000 -u 10000 chart
+
+COPY _output/$RELEASE_BRANCH/dependencies/$TARGETOS-$TARGETARCH/eksa/helm/chartmuseum/chartm /home/chart
+COPY ../harbor/make/photon/chartserver/docker-entrypoint.sh /home/chart/
```
jaxesn, line 14: We dont do this much but i think it might be nicer if we copy these files into output_dir somewhere to avoid this.

projects/goharbor/harbor/docker/linux/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY ../harbor/make/photon/common/install_cert.sh /home/chart/
+COPY _output/LICENSES /LICENSES
+COPY ATTRIBUTION.txt /ATTRIBUTION.txt
+
+RUN chown -R chart:chart /etc/pki/tls/certs \
```
jaxesn, line 19: i think we for the stuff in home/chart you could add the --chown the COPY command above.

projects/goharbor/harbor/docker/linux/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && chmod u+x /home/chart/docker-entrypoint.sh \
+    && chmod u+x /home/chart/install_cert.sh
+
+USER chart
+
+WORKDIR /home/chart
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
+
+VOLUME ["/chart_storage"]
```
jaxesn, line 31: We never really add these calls, not sure if it is necessary or not since kubernetes is going to override them anyway?

projects/goharbor/harbor/docker/linux/Dockerfile
```diff
@@ -0,0 +1,33 @@
+ARG BASE_IMAGE # https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-base
+FROM $BASE_IMAGE
+
+ARG RELEASE_BRANCH
+ARG TARGETARCH
+ARG TARGETOS
+
+RUN yum install -y shadow-utils >> /dev/null \
```
jaxesn, line 8: I want to talk thru this part, its kind of dependent on what base image we use.

projects/goharbor/harbor/docker/linux/Dockerfile
```diff
@@ -0,0 +1,33 @@
+ARG BASE_IMAGE # https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-base
```
jaxesn, line 1: What is this image used for?  It looks like it has the chart in it?  Is that a cli?

jaxesn (changes requested):
Good start!  I havent look at chart museum yet, but I think this is probably enough to get started on the changes.  We will need to add prow jobs and make the CDK changes for these.

projects/helm/chartmuseum/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+BASE_IMAGE_NAME?=eks-distro-minimal-base
+SOURCE_PATTERNS=./cmd/chartmuseum
+
+BINARY_TARGET_FILES=chartm
+
+HAS_S3_ARTIFACTS=true
+IMAGE_NAMES=
+
+PRECOMPILE_PREPARE=harbor_patch
```
jaxesn, line 16: Same comment as above, I think you can do this by adding a pre-req to the binary target

projects/helm/chartmuseum/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+HARBOR_VERSION=v2.4.1
+
+harbor_patch:
+	cd $(PROJECT_ROOT) \
+	&& git clone https://github.com/goharbor/harbor.git harbor \
+	&& git -C harbor checkout -f $(HARBOR_VERSION) \
+	&& cd $(REPO) && git apply ../harbor/make/photon/chartserver/redis.patch \
```
jaxesn, line 24: I think i would instead pull this into our repo in a patches directory here and let it apply like we do other patches to avoid this part.  A bit of duplication, but I think that would be cleaner and avoid additional clones that we may have to worry about when we are thinking about pulling everything from codecommit in codebuild.

jaxesn (changes requested):
Couple comments on museum

projects/goharbor/harbor/Makefile
vincentni: Harbor uses scripts heavily for its service initialization and configuration. I don't think it is possible to implement those complicated logics (conditional & loops) with vanilla commands in Dockerfile. Yes, bash is the major reason I go for the standard image. Besides that, a quite few packages are also needed in the runtime - tzdata, shadow-utils, cronie, rsyslog, logrotate, tar, gzip, sudo, etc.

projects/goharbor/harbor/Makefile
vincentni: go.mod file resides in src folder.

projects/goharbor/harbor/Makefile
vincentni: oh yeah, BASE_IMAGE_NAME can go as it is default to that. I don't think REPO_SUBPATH and DOCKERFILE_FOLDER can go. Am I missing something?

projects/goharbor/harbor/Makefile
vincentni: That is right if it has only one image. I name it that way because we are expecting more image components to be added.

projects/goharbor/harbor/Makefile
vincentni: it is used in DOCKERFILEPATH_CHART_SERVER, but yeah I can fold this in.

projects/goharbor/harbor/Makefile
vincentni: swagger code generation needs to be done after repo checkout and right before go mod download otherwise the compilation fails. I don't think it achieves what I want with the suggested change.

projects/goharbor/harbor/.gitignore
```diff
@@ -0,0 +1 @@
+harbor
```
TerryHowe, line 1:
```suggestion
/harbor
```

projects/helm/chartmuseum/README.md
```diff
@@ -0,0 +1,5 @@
+## **chartmuseum**
+![Version](https://img.shields.io/badge/version-v0.3.7-blue)
+![Build Status](https://codebuild.us-west-2.amazonaws.com/badges?uuid=eyJlbmNyeXB0ZWREYXRhIjoiZVJQQTc0Vk8rcHlMR0hOYnllRGNmV0NsQTNLNGFaS2hLME1MUmgwYkxpVUFoL0V0WHZzbXVCV1owQ0FUTlF6RHg1WXhWRXZLRzNwN2d2LzZGUVJvZ0pRPSIsIml2UGFyYW1ldGVyU3BlYyI6Im9jQmZMa216aHZpYmdrWDYiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D&branch=main)
+
+The [chartmuseum project](https://github.com/helm/chartmuseum) is an open-source Helm Chart Repository server written in Go (Golang), with support for cloud storage backends, including Google Cloud Storage, Amazon S3, Microsoft Azure Blob Storage, Alibaba Cloud OSS Storage, Openstack Object Storage, Oracle Cloud Infrastructure Object Storage, Baidu Cloud BOS Storage, Tencent Cloud Object Storage, Netease Cloud NOS Storage, DigitalOcean Spaces, Minio, and etcd.
```
TerryHowe, line 5: Unless Harbor doesn't work without this thing running, we really don't need a helm chart repo for our initial release. Currently, we store our helm charts in a container image registry.

projects/helm/chartmuseum/.gitignore
```diff
@@ -0,0 +1 @@
+chartmuseum
```
TerryHowe, line 1:
```suggestion
/chartmuseum
```

projects/goharbor/harbor/Makefile
jaxesn: What other images are lookign to build in this project?  Unless we are adding them now, I'd prefer we treat this as standard as possible

projects/goharbor/harbor/docker/linux/harbor-chartserver/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && useradd --no-log-init -m -g 10000 -u 10000 chart
+
+COPY --chown chart:chart _output/$RELEASE_BRANCH/dependencies/$TARGETOS-$TARGETARCH/eksa/helm/chartmuseum/chartm /home/chart
+COPY --chown chart:chart _output/make/photon/chartserver/docker-entrypoint.sh /home/chart/
+COPY --chown chart:chart _output/make/photon/common/install_cert.sh /home/chart/
+COPY --chown chart:chart _output/LICENSES /LICENSES
+COPY --chown chart:chart ATTRIBUTION.txt /ATTRIBUTION.txt
+
+RUN chown -R chart:chart /etc/pki/tls/certs \
+    && chown -R chart:chart /home/chart \
```
jaxesn, line 20: I'm not sure i see the need for these 4 lines, do the scripts not already have the +x flag? isnt the home/chart dir already owned by chart from the copy call?  

Similar question to the certs dir, is not world readable?

projects/helm/chartmuseum/Makefile
```diff
@@ -0,0 +1,24 @@
+BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
+GIT_TAG?=$(shell cat GIT_TAG)
+GOLANG_VERSION?="1.17"
+
+REPO=chartmuseum
+REPO_OWNER=helm
+
+BASE_IMAGE_NAME?=eks-distro-minimal-base
```
jaxesn, line 8: You can drop this since we arent building any images.

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+harbor-chartserver-useradd/images/export: IMAGE_USERADD_USER_NAME=chart IMAGE_USERADD_USER_ID=10000
+harbor-chartserver-useradd/images/export: scripts_prepare
+
+HARBOR_BASE_NAMESPACE=goharbor
+IMAGE_BUILD_ARGS=HARBOR_BASE_NAMESPACE
+
+swagger_prepare:
+	make -C $(REPO) gen_apis
+
+scripts_prepare:
```
jaxesn, line 32: Lets make this a real path target.  What we using do is define it as a var up top kind of like this

```
HARBOR_SCRIPTS_TARGET=$(OUTPUT_DIR)/make/photon/common/install_cert.sh
....
$(HARBOR_SCRIPTS_TARGET):
  
  
...


```

projects/goharbor/harbor/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+$(call IMAGE_TARGETS_FOR_NAME, harbor-chartserver): harbor-chartserver-useradd/images/export
+
+harbor-chartserver-useradd/images/export: IMAGE_USERADD_USER_NAME=chart IMAGE_USERADD_USER_ID=10000
+harbor-chartserver-useradd/images/export: scripts_prepare
+
+HARBOR_BASE_NAMESPACE=goharbor
+IMAGE_BUILD_ARGS=HARBOR_BASE_NAMESPACE
+
+swagger_prepare:
```
jaxesn, line 29: Like the comment below, if its possible to point a generated file for this target that would be nicer.

jaxesn (changes requested):
more review

projects/helm/chartmuseum/patches/redis.patch
```diff
@@ -0,0 +1,77 @@
+diff --git a/cmd/chartmuseum/main.go b/cmd/chartmuseum/main.go
```
TerryHowe, line 1: I'm not sure where this patch came from, but should probably use `git format-patch` to create it and have a link to upstream issue/PR for it if there is something. I don't see a PR to add this feature in helm/chartmuseum

vincentni: It is the patch from harbor for this component. https://github.com/goharbor/harbor/blob/v2.4.1/make/photon/chartserver/redis.patch

vincentni: As suggested by Jackson, instead of checking out harbor source code and applying patches during the build time, it is preferred to copy those patches to the patches folder.

PR #375, opened 2022-01-28
Avoid creating source directories on destination during rsync copy

abhay-krishna:
Avoid creating source directories recursively on destination when using `rsync`.


pokearu (approved):
/lgtm

PR #378, opened 2022-01-28
Use absolute path for bottlerocket download directory

abhay-krishna:
Not setting this makes it relative to the root directory.


pokearu (approved):
/lgtm

PR #382, opened 2022-01-29
Move raw image build packer log one level up

abhay-krishna:
The packer log parent folder must already exist.


pokearu (approved):
/lgtm

PR #386, opened 2022-02-01
Add supported release branches file

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Decouple supported release branches from the versions of Bottlerocket ovas that we support. This would allow us to start testing with Ubuntu support first.


jaxesn (approved):
Great idea

/lgtm

PR #393, opened 2022-02-03
Clone ipmitool repo outside docker build stage

abhay-krishna:
Cloning outside Docker build since we don't have AWS access inside docker build (to clone from CodeCommit).


pokearu (approved):
/lgtm

🤞

PR #396, opened 2022-02-04
Align CLI Tools and diagnostic collector tags with EKS-A CLI

abhay-krishna:
Since these are projects that we define ourselves for use with the CLI, we have some flexibility with the tags we assign to them. Tagging them with the same version as the EKS-A CLI seems reasonable.


TerryHowe (approved):
Might want a note on this in the release runbook

/lgtm

PR #399, opened 2022-02-07
kind: fix high cpu agetty processes

jaxesn:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1159

*Description of changes:*
Per the ticket above when creating CAPD clusters there is a `agetty` process running which is attached (?) to the tty1 interface, but since it's in a container it's not really able to attach(?).  According to an (old) [issue](https://github.com/moby/moby/issues/4040) logged against docker this has to do something with privileged containers using `init` as their entrypoint, which the kind image def does.  Looking at some of the attached tickets there appears to be two approaches to solving the issue:

- masking, or removing, the getty@ttty1 service in /etc/systemd/system
- replacing `/bin/agetty` with `/bin/true`

This [issue](https://github.com/voxpupuli/beaker-hostgenerator/pull/138) went with the replacing option citing issues with other ttys and processes not exiting with containers exit.  I was not able to reproduce the latter and am not totally sure if/when a container might be launched on a host with multiple ttys.

When disabling the service, there is still a agetty process running that looks like `/sbin/agetty --noclear --keep-baud console 115200,38400,9600 vt220`.  This one does not seem to eat all the cpu and does go away correctly when the container exits.

I went with the first option of disabling the service when building the kind base image.  This feels safer...  If anyone knows more and wants to push the other way and/or if we see other issues in the future we could go with option 2.

Testing:

- Built kind image locally and create a CAPD with it and only saw the one agetty process per container and cpu usage was fine.

For reference, when running docker on mac, to attach to the docker host/vm, I follow this: https://gist.github.com/BretFisher/5e1a0c7bcca4c735e716abf62afad389 specifically

`docker run -it --rm --privileged --pid=host justincormack/nsenter1`

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -1,6 +1,6 @@
 # Disable built-in rules and variables
 MAKEFLAGS+=--no-builtin-rules --warn-undefined-variables
-SHELL=/bin/bash
+SHELL=bash
 .SHELLFLAGS:=-eu -o pipefail -c
 .SUFFIXES:
 
--- a/projects/kubernetes-sigs/kind/patches/0001-Switch-to-AL2-base-image-for-node-image.patch
+++ b/projects/kubernetes-sigs/kind/patches/0001-Switch-to-AL2-base-image-for-node-image.patch
@@ -1,16 +1,16 @@
-From 0d2af05bf029506e0e563d9e6cb515963d1333e9 Mon Sep 17 00:00:00 2001
+From 7476c8c48137973bec9c782b9518d99e2284bbb6 Mon Sep 17 00:00:00 2001
 From: Jackson West <jgw@amazon.com>
 Date: Thu, 22 Jul 2021 12:59:46 -0500
 Subject: [PATCH] Switch to AL2 base image for node image
 
 Signed-off-by: Jackson West <jgw@amazon.com>
 ---
- images/base/Dockerfile                        | 141 ++++++++----------
+ images/base/Dockerfile                        | 145 +++++++++---------
  images/base/files/usr/local/bin/clean-install |   7 +-
- 2 files changed, 70 insertions(+), 78 deletions(-)
+ 2 files changed, 73 insertions(+), 79 deletions(-)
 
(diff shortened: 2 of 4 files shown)


Common.mk
```diff
@@ -1,6 +1,6 @@
 # Disable built-in rules and variables
 MAKEFLAGS+=--no-builtin-rules --warn-undefined-variables
-SHELL=/bin/bash
+SHELL=bash
```
jaxesn, line 3: After updating my mac to monterey, make started using the mac default bash at /bin/bash which doesn't support `declare -A` which we use in various scripts.  I have bash 5 installed via brew.  I am not sure if at one point before the upgrade i had bin/bash symlinked differently or what.  I *think* this change should be ok and still work in the al2 builder base, but well see in prow.

projects/kubernetes-sigs/kind/patches/0001-Switch-to-AL2-base-image-for-node-image.patch
```diff
@@ -70,24 +70,28 @@ index ec8fcd7c..436f9854 100644
...
      && rm -f /lib/systemd/system/multi-user.target.wants/* \
-     && rm -f /etc/systemd/system/*.wants/* \
-@@ -111,44 +96,51 @@ RUN echo "Installing Packages ..." \
+-    && rm -f /etc/systemd/system/*.wants/* \
++    && rm -vf /etc/systemd/system/*.wants/* \
+     && rm -f /lib/systemd/system/local-fs.target.wants/* \
+     && rm -f /lib/systemd/system/sockets.target.wants/*udev* \
      && rm -f /lib/systemd/system/sockets.target.wants/*initctl* \
      && rm -f /lib/systemd/system/basic.target.wants/* \
++    # avoid runaway agetty processes most likely due to al2 being based on older centos 7
```
jaxesn, line 93: These are the only 2 lines i added.

TerryHowe (approved):
/lgtm

PR #413, opened 2022-02-11
Helm Charts for Harbor Addon

vincentni:
*Issue #, if available:*

*Description of changes:*


jaxesn (commented):
This looks fine to me.  My main question is if we want to do it this way or more like we did cert-manager, where the main cert-manager build clones the additional helm chart repo and builds it there instead of it being its own project.  I dont have a strong opinion either way.

We need the prow jobs here too.

PR #414, opened 2022-02-11
Updating GIT_TAG for kube-vip

pokearu:
*Description of changes:*
For Baremetal provider we required dynamic interface mapping from kube-vip. Which was added in 
[289](https://github.com/kube-vip/kube-vip/pull/289)
Hence bumping up kube-vip release.


jaxesn (commented):
This is the kind of thing we need to be somewhat careful about.  Since this is used by both capv and now capt, we need to ensure the version is compatible.  @mrajashree can confirm, but i suspect we didnt update this before because capv is still using the 3.7 version upstream.  I also suspect it doesnt matter if we get ahead of capv here, but it should be a known choice we are making.

mrajashree:
@jaxesn that's right CAPV v1.0.1 tag which we're using is using an even [older version 0.3.5](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/v1.0.1/packaging/flavorgen/flavors/generators.go#L254) of kube-vip that's why we didn't upgrade it. But it looks like it should be fine to use the latest version for capv

PR #416, opened 2022-02-12
Revert "Updating GIT_TAG for kube-vip (#414)"

pokearu:
This reverts commit d2ceb47a9c7ac8c686ef4f3fa016ddc5fece53e5.

*Description of changes:*
Reverting version upgrade. This version is causing CAPV provider to error.

***Needs further investigation.***


UPSTREAM_PROJECTS.yaml
```diff
@@ -76,7 +76,6 @@ projects:
     repos:
       - name: cloud-provider-vsphere
         versions:
-          - tag: v1.22.5
```
abhay-krishna, line 79: nit: this can stay

PR #424, opened 2022-02-15
Add CAPT patch to remove requirement on machine.Spec.Version

abhinavmpandey08:
*Issue #, if available:*

*Description of changes:*
The CAPI machine crds for external etcd currently do not set `machine.Spec.Version` but CAPT makes it a required field. Etcd machines fail to reconcile as a result.

This patch removes this requirement allowing us to use external etcd with CAPT.


jaxesn (approved):
/lgtm

PR #431, opened 2022-02-16
pbnj: use destdir when make installing impitool

jaxesn:
*Issue #, if available:*

*Description of changes:*

This will make sure everything make install puts in place, is in the resulting image, including the license.


abhay-krishna (approved):
/lgtm

PR #432, opened 2022-02-16
Creating ARTIFACTS_PATH before cp

pokearu:
*Description of changes:*
Since we set `SIMPLE_CREATE_TARBALLS=false` we need to create the artifacts path to copy the final OSIE files.


abhay-krishna (approved):
/lgtm

PR #433, opened 2022-02-16
Make registry certificate hosting URL and tink-worker image

ptrivedi:
This patch makes registry certificate hosting URL and tink-worker image name:tag configurable

* If you are using a registry with a self-signed certificate, you can specify
  a custom URL of where the certificate is hosted. On boots, use REGISTRY_CERT_URL
  environment variable to specify the URL to the registry CA certificate. Boots
  passes this over to hook via command-line parameter 'registry_cert_url'

* In order to specify a custom name:tag for the tink-worker image instead of
  defaulting to the hardcoded default of 'tink-worker:latest', use the
  TINK_WORKER_IMAGE environment variable setting in boots. Boots passes this over
  to hook via command-line parameter 'tink_worker_image'

*Issue #, if available:*

*Description of changes:*


projects/tinkerbell/hook/patches/0001-Make-registry-certificate-hosting-URL-and-tink-worke.patch
```diff
@@ -0,0 +1,125 @@
...
+@@ -39,7 +40,13 @@ func main() {
+ 		panic(err)
+ 	}
+ 	// Download the configuration
+-	err = DownloadFile(path+"ca.crt", cfg.baseURL+"/ca.pem")
++	var baseCertURL string
++	if (len(cfg.registryCertURL) > 0) {
++		baseCertURL = cfg.registryCertURL
++	} else {
++		baseCertURL = cfg.baseURL
```
pokearu, line 107: So baseURL here would be tinkerbell nginx IP right? 
So if we don't use a `local/private` registry and we use something like public ECR to pull `tink-worker` we technically don't need any certs right?
In this case what file would be downloaded here?

ptrivedi: This comment makes me want to put docker cert directory creation etc. under a conditional based on whether registry certificate path url optional parameter is present

PR #451, opened 2022-02-18
Add tinkerbell stack to tinkerbell project

abhinavmpandey08:
*Issue #, if available:*

*Description of changes:*


abhinavmpandey08:
> There are a number of hardcoded image tag/s3 locations here. That seems like it could be an issue? Is there something at runtime that replaces these?

No, there isn't. The current expectation is that when we release this, we will change all these tags to release tags manually atleast for beta. Does that sound reasonable?

pokearu:
> There are a number of hardcoded image tag/s3 locations here. That seems like it could be an issue? Is there something at runtime that replaces these?

Yea like this is a temp thing. We plan to get rid of the whole docker-compose pretty soon. 
We just wanted to make changes to the compose file in a common place.

abhinavmpandey08:
> This looks good to me. Do we wanna name it tinkerbell-stack or sandbox? based on the upstream project

i don't feel strongly about either name. I can rename it to sandbox

pokearu (approved):
/lgtm

PR #460, opened 2022-02-22
remove duplicated validate-checksums target

jaxesn:
*Issue #, if available:*

*Description of changes:*

I cant remember or figure out why there were two targets for checksums.  I remember there being some weirdness around updating checksums but that has long since been fixed/removed.  Im pretty sure we can clean this up now since they were the same target, minus the if which should def be there.


TerryHowe (approved):
/lgtm

PR #367, opened 2022-01-26
Harbor addon

vincentni:
*Description of changes:*
the following images are built:
harbor/harbor-core harbor/harbor-db harbor/harbor-exporter harbor/harbor-jobservice harbor/harbor-log harbor/harbor-nginx harbor/harbor-portal harbor/harbor-redis harbor/harbor-registry harbor/harbor-registryctl

the remaining images to be built:
harbor/harbor-chart-server harbor/harbor-notary-signer harbor/harbor-notary-server harbor/harbor-trivy-adapter

PR #466, opened 2022-02-22
add distribution project

vincentni:
*Issue #, if available:*

*Description of changes:*


projects/distribution/distribution/GIT_TAG
```diff
@@ -0,0 +1 @@
+v2.7.1
```
abhay-krishna, line 1: We have received Depedabot alerts regarding Distribution project tags below `v2.8.0`? Can we build off [that tag](https://github.com/distribution/distribution/releases/tag/v2.8.0) instead, considering they've added CVE patches?

TerryHowe: Is this the same project at docker/distribution? If so, I think it needs to be `v2.8.0-beta.1`

abhay-krishna: Yes it's the same project, but the `beta.1` tag is just a pre-release tag (that's why the release changelogs for `v2.8.0-beta.1` and the actual release tag `v2.8.0` are identical). Also there is only one commit diff between these two, that too in a README. 

https://github.com/distribution/distribution/compare/v2.8.0-beta.1..v2.8.0

I don't think it would hurt to just use the latest, but I may be wrong,

vincentni: Let's get this through to unblock another PR (https://github.com/aws/eks-anywhere-build-tooling/pull/367). I'll make another PR to deal with CVE.

projects/distribution/distribution/README.md
```diff
@@ -0,0 +1,5 @@
+## **distribution**
+![Version](https://img.shields.io/badge/version-v0.3.7-blue)
```
abhay-krishna, line 2:
```suggestion
![Version](https://img.shields.io/badge/version-v2.8.0-blue)
```

projects/distribution/distribution/Makefile
```diff
@@ -0,0 +1,45 @@
...
+GOLANG_VERSION?="1.17"
+
+REPO=distribution
+REPO_OWNER=distribution
+
+BINARY_TARGET_FILES=registry
+SOURCE_PATTERNS=./cmd/registry
+
+VERSION=$(shell git -C $(REPO) describe --match 'v[0-9]*' --dirty='.m' --always)
+REVISION=$(shell git -C $(REPO) rev-parse HEAD)$(shell if ! git -C $(REPO) diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
```
abhay-krishna, line 12: Do these commands return the same output for a given tag, i.e., do they affect build reproducibility?

projects/distribution/distribution/README.md
```diff
@@ -0,0 +1,5 @@
+## **distribution**
+![Version](https://img.shields.io/badge/version-v2.8.0-blue)
+![Build Status](https://codebuild.us-west-2.amazonaws.com/badges?uuid=eyJlbmNyeXB0ZWREYXRhIjoiZVJQQTc0Vk8rcHlMR0hOYnllRGNmV0NsQTNLNGFaS2hLME1MUmgwYkxpVUFoL0V0WHZzbXVCV1owQ0FUTlF6RHg1WXhWRXZLRzNwN2d2LzZGUVJvZ0pRPSIsIml2UGFyYW1ldGVyU3BlYyI6Im9jQmZMa216aHZpYmdrWDYiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D&branch=main)
```
abhay-krishna, line 3: Is this the correct build status for the distribution project or from another readme? Possibly kubevip?

projects/distribution/distribution/Makefile
vincentni: no, they won't

PR #470, opened 2022-02-22
Project to build the Snow EKS-A Admin AMI

g-gaston:
This is part of https://github.com/aws/eks-anywhere/issues/1166

*Description of changes:*
This uses AWS image builder to build the Admin AMI for Snow. Most of the explanation about how it works is in the README. But in summary:
* One command setups the infrastructure. We should run this every time there is a change in this repo. Gotcha: this will successfully create new versions of the components but it won't update anything else.
* Second command builds the AMI. In order to do this: it creates a new image recipe with the input params (eks-a version and release manifest url), updates the image pipeline to use it and then runs the pipeline.

Some other notes:
* The AMI is incomplete right now. We need to modify some of the eks-a commands (import, download, etc.) and add new modules to the recipe to run them. I will start working on that once we have this up and running. I wrote this so is very easy to add new modules.
* I think this is pattern is wrong. If you look at the code, I ended up implementing a poor/hacky of something that resembles and incomplete version of CDK. Hence, I think we should use CDK. This will most probably present other challenges (like permissions, handle drift when the pipeline is updated...). So at least this gives us something working while we figure those out.
* This has a huge lack of tests. Since I don't think we should continue with this pattern for long (plus the whole thing takes a couple minutes to run locally), I thought it would be wasteful to write more tests. But please let me know if you think otherwise.
* ~It's missing a `Makefile`. I will add it if everyone is ok with following this approach.~ Thanks to @jaxesn 


g-gaston:
Closing in favor of #710

PR #472, opened 2022-02-23
Add common buildspec for all projects

abhay-krishna:
Each project folder has a buildspec, but they're all identical. So we should replace them with a single one at root. This change will be done as a blue-green deployment.


TerryHowe (approved):
/lgtm

PR #481, opened 2022-02-24
Remove upload-artifacts phase from buildspec

abhay-krishna:
This is handled by the `release` target now and doesn't need to be a separate phase.


jaxesn (approved):
/lgtm

PR #504, opened 2022-03-02
Add etcdadm patch to accept multiple endpoints

mrajashree:
*Issue #, if available:*

*Description of changes:*
During etcd cluster creation, etcdadm first creates a single member etcd cluster through the `etcdadm init` command, and then the rest of the members can join the cluster using the `etcdadm join <endpoint>` command.
If this first member gets deleted, we need to decide on a new endpoint to be used for the join command. Instead if we pass in all etcd members to the join command we won't need to check if one of them is unavailable.
This endpoint is only used within etcdadm to generate the etcd client. Etcd client can be generated for multiple endpoints. This commit modifies the etcd client generation to use as many endpoints are passed in. 

*Testing*
Built etcdadm binary, and used it through the etcdadm-bootstrap-provider's logic to [install a different etcdadm binary instead of using the baked in etcdadm](https://github.com/mrajashree/etcdadm-bootstrap-provider/blob/v1beta1/controllers/etcdadmconfig_controller.go#L247). Verified that the init and join commands worked and the logs contained the addition log added in this PR.
Tested EKS-A cluster creation, and tested deletion of the machine/VM corresponding to the first etcd leader and saw it gets replaced.


g-gaston (approved):
/lgtm

PR #518, opened 2022-03-04
cilium: fix manifest generationt to use proper tag

jaxesn:
*Issue #, if available:*

*Description of changes:*

*NOTE*: the cli will soon use the helm chart directly to install so this manifest generation will go away.

The cilium manifest generation uses a values yaml override file when generating the final manifest from helm.  There was a hardcoded image tag in that file that did not match what was in the GIT_TAG file.  I suspect, but did not verify, this means that on install the cli was still installing the older image, but during upgrade since it uses the helm chart it would install the newer GIT_TAG.  I changed the manifest generation to use the helm chart from public ecr, which the cli is using during upgrade and to use the GIT_TAG value as the image tag.  This will ensure we don't have two places to update the version.

Also, I bumped to the latest 1.9 patch release from isovalent.


g-gaston (approved):
/lgtm

PR #525, opened 2022-03-07
Increase eks-d AMI sice to 25GB

g-gaston:
*Description of changes:*
Until we don't have the option to dynamically set the VM disk size, this
is the only way to have enough disk space to download container images.

### DIFF
--- a/projects/kubernetes-sigs/image-builder/packer/ami/packer.json
+++ b/projects/kubernetes-sigs/image-builder/packer/ami/packer.json
@@ -5,5 +5,6 @@
   "snapshot_groups": "",
   "custom_role": "true",
   "custom_role_names": "{{env `BUILDER_ROOT`}}/ansible/roles/load_additional_files",
-  "ansible_extra_vars": "@{{env `BUILDER_ROOT`}}/packer/ami/ansible_extra_vars.yaml"
+  "ansible_extra_vars": "@{{env `BUILDER_ROOT`}}/packer/ami/ansible_extra_vars.yaml",
+  "volume_size": "25"
 }


projects/kubernetes-sigs/image-builder/packer/ami/packer.json
```diff
@@ -5,5 +5,6 @@
   "snapshot_groups": "",
   "custom_role": "true",
   "custom_role_names": "{{env `BUILDER_ROOT`}}/ansible/roles/load_additional_files",
-  "ansible_extra_vars": "@{{env `BUILDER_ROOT`}}/packer/ami/ansible_extra_vars.yaml"
+  "ansible_extra_vars": "@{{env `BUILDER_ROOT`}}/packer/ami/ansible_extra_vars.yaml",
+  "volume_size": "30"
```
vignesh-goutham, line 9: Correct me if Im wrong, doesn't upstream set the default size to be 25G? Maybe not for ami packer config.

In either case, do we want to stick with 25G across all providers? I'm down for ami to be 30 if that's something we really want.

g-gaston: The default size for this AMI was 8GB. Tbh 30GB was kind of arbitrary. I'm ok with 25 is that is more common.

vignesh-goutham: yea lets do 25, just to be consistent across all providers.

g-gaston: Done

vignesh-goutham (approved):
/lgtm

PR #529, opened 2022-03-08
capv: upgrade to 1.1.1

jaxesn:
*Issue #, if available:*

*Description of changes:*

This starts the update process for capv to 1.1.0.  There isnt a whole lot interesting here, except for the patches changes.  @vignesh-goutham and @mrajashree have upstreamed a couple of the old patches so I think all we need to keep is where we enable BR user data.

As far as the cluster-template and infra-components manifests.

The only change on the cluster-template is the kube-vip upgrade which @abhinavmpandey08 has been working on separately.

In the infra-components, there is a new field `additionalDisksGiB` which @vignesh-goutham said we shouldnt need to set.  The other changes deal with the kcp roles/bindings which do not appear to be an issue.

Holding and handing off to @abhinavmpandey08 to merge/finish when ready

/hold  


vignesh-goutham (approved):
/lgtm

PR #533, opened 2022-03-09
Add support for loading userdata from secretsmanager

johnsonjnpan:
*Description of changes:*

This change adds the ability for the bootstrap container to load userdata from AWS SecretsManager. 
Instead of embedding sensitive data directly inside userdata. We can pass a special crafted userdata (aka. ExternalUserData) that contains a reference to some secret(s) in AWS SecretsManager.

Sample ExternalUserData:
```
#cloud-config
user_data_type: "ExternalUserData"

user_data_source:
  provider: secret-manager
  prefix: <prefix-placeholder>
  chunks: <chunks-placeholder>
```


projects/aws/bottlerocket-bootstrap/pkg/service/secretsmanager.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
+)
+
+var secretsManagerService SecretsManagerService
```
vignesh-goutham, line 11: Instead of a global service like this, can we re-architect this to follow golang recommended way using new(). Something like this
```
type SecretsManager struct {
    BaseClient *secretsmanager.Client
}

func NewSecretsManager() {
    secretsManager = new(SecretsManager)
    secretsManager.blah = foo
    return secretsManager...
}

func (sm *SecretsManager) DeleteSecret(...) {}
```
You can keep the interface.

projects/aws/bottlerocket-bootstrap/pkg/utils/userdata.go
```diff
@@ -24,28 +27,89 @@ type WriteFile struct {
 	Content     string
 }
 
+type UserDataSource struct {
+	Provider string
+	Prefix   string
+	Chunks   int
+}
+
+type ExternalUserData struct {
```
vignesh-goutham, line 36: Looks like this does not house the userdata itself and only the source, can we merge the above struct with this, and rename this one to something like `ExternalUserDataSource`?

projects/aws/bottlerocket-bootstrap/pkg/utils/userdata.go
```diff
@@ -24,28 +27,89 @@ type WriteFile struct {
 	Content     string
 }
 
+type UserDataSource struct {
+	Provider string
```
vignesh-goutham, line 31: Would we have other provider than secrets-manager?

projects/aws/bottlerocket-bootstrap/pkg/utils/userdata.go
```diff
@@ -24,28 +27,89 @@ type WriteFile struct {
...
+	Prefix   string
+	Chunks   int
+}
+
+type ExternalUserData struct {
+	UserDataType   string         `yaml:"user_data_type"`
```
vignesh-goutham, line 37: This looks like this will always be "external" right? Should we get rid of this?

vignesh-goutham (commented):
Looks good overall, have few questions below.

projects/aws/bottlerocket-bootstrap/Makefile
```diff
@@ -41,7 +41,8 @@ include $(BASE_DIRECTORY)/Common.mk
 $(REPO):
 	@mkdir $@
 
-$(call IMAGE_TARGETS_FOR_NAME, bottlerocket-bootstrap): $(call FULL_FETCH_BINARIES_TARGETS, $(FETCH_BINARIES_TARGETS))
+
+$(call IMAGE_TARGETS_FOR_NAME, bottlerocket-bootstrap): unit-test $(call FULL_FETCH_BINARIES_TARGETS, $(FETCH_BINARIES_TARGETS))
```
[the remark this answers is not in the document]
johnsonjnpan: I found that `build` also runs `local-images` target. To enable unit tests for both of them, I'll add

projects/aws/bottlerocket-bootstrap/pkg/utils/userdata.go
johnsonjnpan: Currently secrets-manager is the only one we support. I leave it there because CAPI also support storing user data using SSM parameter store, (even through we are not using it).
I'm open to either keep/remove that.

projects/aws/bottlerocket-bootstrap/pkg/service/secretsmanager.go
johnsonjnpan: Yes, will do

projects/aws/bottlerocket-bootstrap/Makefile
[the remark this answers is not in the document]
johnsonjnpan: Added `unit-test` in `build`

johnsonjnpan:
I sent out a few new commits and refactored the "ExternalUserData" concept to a more flatten one.
Before, I'm thinking the following structure:
```
ProcessUserData()
{
    if(userdata.type=="ExternalUserData)
    {
        // parseUserData as ExternalUserData
        if (userdata.provider="secrets-manager")
        {
            // handles secrets-manager
        }
    }
    else
    {
        // handles normal userdata
    }
}
```
After refactoring, we are throwing away the concetp of external user data:
```
ProcessUserData()
{
    if(userdata.type=="AWSSecretsManager)
    {
        //parse userdata as AWSSecretsManagerUserData
        // handles AWSSecretsManagerUserData
    }
    else
    {
        // handles normal userdata
    }
}

```

This should makes the handling of userdata a lot cleaner

PR #537, opened 2022-03-09
Consume prod release of EKS-D 1.22

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
The production release is out for EKS-D now, so changing links to consume that for our bundles.

### DIFF
--- a/EKSD_LATEST_RELEASES
+++ b/EKSD_LATEST_RELEASES
@@ -12,7 +12,6 @@ releases:
   number: 9
   kubeVersion: v1.21.5
 - branch: 1-22
-  number: 6
+  number: 1
   kubeVersion: v1.22.6
-  dev: true
 latest: 1-21


EKSD_LATEST_RELEASES
```diff
@@ -12,7 +12,6 @@ releases:
   number: 9
   kubeVersion: v1.21.5
 - branch: 1-22
-  number: 6
+  number: 1
   kubeVersion: v1.22.6
-  dev: true
 latest: 1-21
```
jaxesn, line 17: Should we bump this now? This basically effects the builds who pull in kubectl into their images and don't care about a specific version.

vivek-koppuru: I think we can start testing with our prod release just to make sure right?

jaxesn: ok, well update this before we ship?

PR #540, opened 2022-03-10
snow: add image build to push to our ecr

jaxesn:
*Issue #, if available:*

*Description of changes:*
We want to push the snow image to our ecr so it matches the others. We are not building it yet, but we can pull and push the one they are producing.

TODO:

- [x] get a real tag from snow, instead of latest

/hold


abhay-krishna (approved):
/lgtm

PR #543, opened 2022-03-11
CAPT patches mutex lock and error return

pokearu:
*Description of changes:*
Adding patches to CAPT for

1. Mutex lock for allocating hardware to machines in controller
2. Temporarily commenting error return for set boot dev

### DIFF
--- a/projects/tinkerbell/cluster-api-provider-tinkerbell/patches/0011-Temporary-comment-error-return.patch
+++ b/projects/tinkerbell/cluster-api-provider-tinkerbell/patches/0011-Temporary-comment-error-return.patch
@@ -1,0 +1,27 @@
+From eaa5811ede0e998454fd3b8be2c1813f9082ad7d Mon Sep 17 00:00:00 2001
+From: Aravind Ramalingam <ramaliar@amazon.com>
+Date: Fri, 11 Mar 2022 10:53:25 -0800
+Subject: [PATCH] Temporary comment error return
+
+---
+ pbnj/controllers/controller.go | 4 +++-
+ 1 file changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/pbnj/controllers/controller.go b/pbnj/controllers/controller.go
+index 591b55f..4ecbfad 100644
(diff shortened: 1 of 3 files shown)


projects/tinkerbell/cluster-api-provider-tinkerbell/patches/0011-Temporary-comment-error-return.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/pbnj/controllers/controller.go
++++ b/pbnj/controllers/controller.go
+@@ -75,7 +75,9 @@ func (r *Reconciler) reconcileNormal(ctx context.Context, bmc *pbnjv1alpha1.BMC)
+ 		if err != nil {
+ 			logger.Error(err, "Failed to set boot device", "BootDevice", bmc.Spec.BootDevice)
+ 
+-			return ctrl.Result{}, fmt.Errorf("failed to set boot device: %s", bmc.Spec.BootDevice) //nolint:goerr113
++			// TODO: Temporary fix to not error out on boot set failure.
++			// This change will be reverted when we have better error propagations
++			// return ctrl.Result{}, fmt.Errorf("failed to set boot device: %s", bmc.Spec.BootDevice) //nolint:goerr113
```
chrisdoherty4, line 21: Is this an attempt to circumvent the boot device problem on Dells?

ptrivedi: We want to power off the server even though the action to set to netboot failed, for now, for bugbash case. It doesn't matter on our servers because we have them set to netboot from BIOS, so powering off is sufficient. But we want to go back and see what we want to do for these servers that don't support set boot device action.

Ideally, we were thinking that we would have CLI repeat some of these checks at the end of cluster deletion. That way we can give a big meaningful warning to the customer saying their hardware did not let us set the boot device to netboot and hence they may need to set it to netboot manually if they want to include the server in the next run of provisioning and cluster creation.

PR #551, opened 2022-03-14
Add bash logging to export image script and remove jq pretty printing

g-gaston:
*Issue #, if available:*

*Description of changes:*
Not sure why the job failed, it was something to do with jq so I suspect it was in line 44. Adding `+x` just in case


abhay-krishna (approved):
/lgtm

PR #554, opened 2022-03-15
Build helm charts with a semantic version.

ewollesen:
Helm 3.8 requires packages to be versioned via a semantic version.

https://github.com/aws/eks-anywhere-packages/issues/31

*Issue #, if available:*

*Description of changes:*


build/lib/helm_push.sh
```diff
@@ -20,13 +20,12 @@ set -o pipefail
 
 IMAGE_REGISTRY="${1?First argument is image registry}"
 HELM_DESTINATION_REPOSITORY="${2?Second argument is helm repository}"
-IMAGE_TAG="${3?Third argument is image tag}"
+VERSION="${3?Third argument is the (semantic) version}"
```
lewisdiamond, line 23: iirc this IMAGE_TAG was actually overridden by what's in GIT_TAG right? So we don't need to keep it and this change is fine.

TerryHowe (approved):
/lgtm

TerryHowe:
```
Error: cannot load values.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"jetstack/cert-manager-cainjector":interface {}(nil)}
```

I ran this manually in codebuild and it worked. Something is wrong in the values.yaml though. It would be interesting to create a PR to just trigger a build in cert-manager to see if it this works

build/lib/helm_push.sh
ewollesen: I don't think so, GIT_TAG was part of what went into IMAGE_TAG. But the only place it was used in this file, as in the generation of a filename (line 28).

TerryHowe:
https://github.com/aws/eks-anywhere-build-tooling/pull/556

Common.mk
```diff
@@ -470,14 +470,14 @@ helm/build: $(OUTPUT_DIR)/ATTRIBUTION.txt
...
-	$(BUILD_LIB)/helm_require.sh $(IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR) $(IMAGE_TAG) $(LATEST) $(HELM_IMAGE_LIST)
+	$(BUILD_LIB)/helm_require.sh $(IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR) $(IMAGE_TAG) $(LATEST) $(VERSION) $(HELM_IMAGE_LIST)
 	$(BUILD_LIB)/helm_replace.sh $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR)
 	$(BUILD_LIB)/helm_build.sh $(OUTPUT_DIR) $(HELM_DESTINATION_REPOSITORY)
 
 # Build helm chart and push to registry defined in IMAGE_REPO.
 .PHONY: helm/push
 helm/push: helm/build
-	$(BUILD_LIB)/helm_push.sh $(IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(IMAGE_TAG) $(OUTPUT_DIR)
+	$(BUILD_LIB)/helm_push.sh $(IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(VERSION) $(OUTPUT_DIR)
```
jaxesn, line 480: Can we generalize version from git_tag or image_tag?  I think this will cause isses for the other projects that build helm charts if there is no version set.

TerryHowe:
It does look like this change broke the presubmit, but PR it worked.

Common.mk
ewollesen: > Can we generalize version from git_tag or image_tag? I think this will cause isses for the other projects that build helm charts if there is no version set.

I think my latest version takes this into account and should hopefully not affect other projects. I've been struggling to build any other projects locally though, so I'm kinda leaning on the presubmit here more than I'd like.

projects/aws/eks-anywhere-test/Makefile
```diff
@@ -1,5 +1,6 @@
 BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
 GIT_TAG?=$(shell cat GIT_TAG)
+HELM_VERSION=$(GIT_TAG:v%=%)+$(shell git rev-parse --short HEAD)
```
jaxesn, line 3: So we basically do this already [here](https://github.com/aws/eks-anywhere-build-tooling/blob/main/Common.mk#L125) for IMAGE_TAG. I wonder if instead of `HELM_VERSION?=$(IMAGE_TAG)-helm` in the Common.mk we could just do `HELM_VERSION?=$(GIT_TAG:v%=%)-$(GIT_HASH)` and then not need to override this?  Do you expect that will be what we do in "most" cases?  I could always continue to be overrideable by individual projects.

ewollesen: I see two possible issues with that approach. The first being, that I think parts of the presubmit tasks are expecting the `-helm` suffix. Those can be updated. However, @TerryHowe mentioned something about how the `-helm` might be what distinguishes the helm chart from the docker image that it installs when stored in a registry? If that's true, and I'm understanding correctly, then it's possible that removing the `-helm` suffix would break things there.

The second part is that we have, for better or for worse, upgraded to helm 3.8, and it requires packages to have a semantic version. `HELM_VERSION?=$(GIT_TAG:v%=%)-$(GIT_HASH)` is not a valid semantic version (the `-` would have to change to a `+`). This too might break some other presubmits, but we can probably update those as well.

TerryHowe:
Alternatively https://github.com/aws/eks-anywhere-build-tooling/pull/585/files

projects/aws/eks-anywhere-test/Makefile
TerryHowe: The -helm is just to disambiguate between container images and helm charts. It kind of sounds like we might move helm charts to their own repo, but that would be a topic for another PR

ewollesen:
/hold

I feel like Terry's is the better solution, unless someone else finds a problem with it, let's go with that.

TerryHowe:
Could probably kill this

PR #569, opened 2022-03-17
Revert "kind: set os-release and include etcd + coredns"

jaxesn:
Reverts aws/eks-anywhere-build-tooling#522

Need to fix the script due to a diff in bash on mac and linux, will resubmit with it fixed.


TerryHowe (approved):
/lgtm

PR #583, opened 2022-03-18
Fix space issues in harbor presubmit

jaxesn:
*Issue #, if available:*

*Description of changes:*

Call buildctl prune between image builds to avoid space issues in presubmit.

Only running prune during presubmit, it would probably be fine to always run it, but going this route to limit the risk of any issues.


Common.mk
```diff
@@ -249,6 +249,7 @@ define BUILDCTL
 		--output type=$(IMAGE_OUTPUT_TYPE),oci-mediatypes=true,\"name=$(IMAGE),$(LATEST_IMAGE)\",$(IMAGE_OUTPUT) \
 		$(if $(filter push=true,$(IMAGE_OUTPUT)),--export-cache type=inline,) \
 		$(foreach IMPORT_CACHE,$(IMAGE_IMPORT_CACHE),--import-cache $(IMPORT_CACHE))
+                buildctl prune --all
```
TerryHowe, line 252: I was thinking about something like that on my mac to build harbor

PR #585, opened 2022-03-18
Change helm charts to use semantic versioning

TerryHowe:


jonahjon (approved):

ewollesen:
Have you tested this? When I tried it locally, it produced a filename without the `v` (because helm package appends the version to the filename), but the `helm_push.sh` script was still expecting the `v`, and so it failed. We might be able to do some shenanigans in `helm_push.sh` to make it work, but that seems like it would be cryptic.

TerryHowe:
I just did a helm/build

TerryHowe:
No idea if there is going to fix upgrade, but it does build and install

ewollesen:
Heh, this is why you make the big bucks, Terry. <3

Would you need to update `helm_require.sh` as well? I'm looking at the metadata/name value in the requires file, and wondering if it shouldn't remove the `v` as well? I'll see if I can make a suggestion.

TerryHowe:
/override harbor-tooling-presubmit
Harbor presubmit fails due to lack of disk space

jaxesn (approved):
/lgtm

PR #587, opened 2022-03-18
Align CLI Tools and diagnostic collector tags with EKS-A CLI

abhay-krishna:
Bumping tags in Tag files and READMEs. In future, it would be nice to have automation around this.


vignesh-goutham (approved):
/lgtm

PR #589, opened 2022-03-19
Modify staging release buildspec to support batch builds

abhay-krishna:
This PR modifies the staging release buildspec to support Codebuild batch builds.

/hold

### DIFF
--- a/release/staging-build.yml
+++ b/release/staging-build.yml
@@ -1,35 +1,192 @@
-version: 0.2
-
-run-as: imagebuilder
-
-env:
-  secrets-manager:
-    GOVC_DATACENTER: "vsphere_ci_beta_connection:vsphere_datacenter"
-    GOVC_URL: "vsphere_ci_beta_connection:vsphere_url"
-    GOVC_USERNAME: "vsphere_ci_beta_connection:vsphere_username"
-    GOVC_PASSWORD: "vsphere_ci_beta_connection:vsphere_password"
-    VSPHERE_CONNECTION_DATA: "vsphere_ci_beta_connection:vsphere_connection_data"
-
-phases:
-  # On AL2, the Cargo build system requires the openssl-devel package
-  # for installing OpenSSL libraries
-  install:
-    run-as: root
-    commands:
-      - yum install -y openssl-devel
-
-  pre_build:
-    run-as: root
-    commands:
-    - ./build/lib/setup.sh
-
-  build:
-    run-as: root
-    commands:
-    - make release-binaries-images
-  
-  post_build:
-    commands:
-    - export HOME="/home/imagebuilder"
-    - make release-ovas
-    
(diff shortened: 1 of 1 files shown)


release/staging-build.yml
```diff
@@ -1,35 +1,245 @@
...
-  
-  post_build:
-    commands:
-    - export HOME="/home/imagebuilder"
-    - make release-ovas
-    
+batch:
+  fast-fail: true
+  build-graph:
+  - identifier: eks_anywhere_diagnostic_collector
```
abhay-krishna, line 4: Using underscores because identifier doesn't support dashes.

release/staging-build.yml
```diff
@@ -1,35 +1,245 @@
...
+      variables:
+       PROJECT_PATH: projects/helm/helm
+       CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/helm.helm
+  - identifier: tink
+    buildspec: buildspec.yml
+    env:
+      variables:
+       PROJECT_PATH: projects/tinkerbell/tink
+       CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/tinkerbell.tink
+  - identifier: bottlerocket_bootstrap
```
jaxesn, line 195: This one depends on etcdadm at least, that may be it but worth a check

release/staging-build.yml
```diff
@@ -1,35 +1,245 @@
...
+      variables:
+       PROJECT_PATH: projects/kubernetes-sigs/cluster-api
+       CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/kubernetes-sigs.vsphere-csi-driver
+  - identifier: cluster_api_provider_aws
+    buildspec: buildspec.yml
+    env:
+      variables:
+       PROJECT_PATH: projects/kubernetes-sigs/cluster-api-provider-aws
+       CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/kubernetes-sigs.cluster-api-provider-aws
+  - identifier: kind
```
jaxesn, line 159: This depends on etcdadm and cri-tools and local path provisioner, double check there aren't more.

jaxesn (changes requested):
Double check through the repo for projects which use fetch artifacts, that should be a decent list of ones which have deps like this.

release/staging-build.yml
abhay-krishna: yup i had them on before, but i think I removed them during testing. Will add them back.

abhay-krishna:
I have included projecs in the buildspec that correspond to non-production-ready providers, like cloudstack and all of the Tinkerbell ones. Do we still need to build them? If not I can remove them.

jaxesn (approved):
/lgtm

PR #599, opened 2022-03-22
Tweak helm chart semantic version generation.

ewollesen:
In commit e09f316 we allowed helm charts to use semantic versioning,
however we made the wrong version semantic. It's not appVersion that
matters, but rather version. This also changes the generated filename.

I also escaped a shell echo that failed when it tried to run on my
machine.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/build/lib/helm_require.sh
+++ b/build/lib/helm_require.sh
@@ -47,7 +47,10 @@ SEDFILE=${OUTPUT_DIR}/helm/sedfile
 export IMAGE_TAG
 export HELM_REGISTRY
 envsubst <helm/sedfile.template >${SEDFILE}
+# Semver requires that our version begin with a digit, so strip the v.
 echo "s,version: v,version: ,g" >>${SEDFILE}
+# Semver requires that we use a + before the git hash to denote build info.
+echo "/^version:/s,-,+," >>${SEDFILE}
 for IMAGE in ${HELM_IMAGE_LIST:-}
 do
   IMAGE_SHASUM=$(${SCRIPT_ROOT}/image_shasum.sh ${HELM_REGISTRY} ${IMAGE} ${LATEST})
--- a/projects/aws/eks-anywhere-test/eks-anywhere-test/Chart.yaml
+++ b/projects/aws/eks-anywhere-test/eks-anywhere-test/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: eks-anywhere-test
 description: EKS Anywhere Packages Test Application
 type: application
-version: {{IMAGE_TAG}}-helm
+version: {{IMAGE_TAG}}
 appVersion: {{IMAGE_TAG}}
--- a/build/lib/helm_push.sh
+++ b/build/lib/helm_push.sh
@@ -23,10 +23,12 @@ HELM_DESTINATION_REPOSITORY="${2?Second argument is helm repository}"
 IMAGE_TAG="${3?Third argument is image tag}"
 OUTPUT_DIR="${4?Fourth arguement is output directory}"
 
+SEMVER_TMP="${IMAGE_TAG#[^0-9]}" # remove any leading non-digits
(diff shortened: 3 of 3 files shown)


projects/aws/eks-anywhere-test/eks-anywhere-test/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: eks-anywhere-test
 description: EKS Anywhere Packages Test Application
 type: application
-version: {{IMAGE_TAG}}-helm
-appVersion: {{IMAGE_TAG}}
+version: {{IMAGE_TAG}}
+appVersion: {{IMAGE_TAG}}-helm
```
lewisdiamond, line 6: Is this correct? Seems like we should just remove `-helm` here, remove the `sed` line that removes it, and not add `-helm` to appVersion.

build/lib/helm_require.sh
```diff
@@ -48,6 +48,8 @@ export IMAGE_TAG
 export HELM_REGISTRY
 envsubst <helm/sedfile.template >${SEDFILE}
 echo "s,version: v,version: ,g" >>${SEDFILE}
```
TerryHowe, line 50: This sed script is case sensitive, so this line above does change `version` and not `appVersion` although the `^` is better

build/lib/helm_require.sh
```diff
@@ -48,6 +48,8 @@ export IMAGE_TAG
...
+echo "/^version:/s,-helm,," >>${SEDFILE}
```
TerryHowe, line 51: Yeh, not sure there is consensus on this, but I think we should get rid of `-helm`

projects/aws/eks-anywhere-test/eks-anywhere-test/Chart.yaml
TerryHowe, line 6: I don't think we should have `-helm` on appVersion because that is user facing

build/lib/helm_require.sh
```diff
@@ -48,6 +48,8 @@ export IMAGE_TAG
...
+echo "/^version:/s,-,+," >>${SEDFILE}
```
TerryHowe, line 52: We might be leaving the `+` for dev builds.

ewollesen: Yes, this isn't removing the `+`, it's changing the `-` to a `+`, so that it's a valid semver.

build/lib/helm_require.sh
ewollesen: Will do

projects/aws/eks-anywhere-test/eks-anywhere-test/Chart.yaml
ewollesen: Will do.

build/lib/helm_require.sh
ewollesen: Yep, it works fine. I'm not changing that behavior, but I can add the `^` if you want me to.

build/lib/helm_require.sh
TerryHowe: I feel like you didn't phrase that right because `-` is valid for semver. You mean if we use the `-` it is ignored by semver, right? I'm just not 100% sure how the production builds will work and I'm not 100% sure 0.1.0+1 is seen by helm has an older version than 0.1.0+2. I was concerned we'd have to change the helm release scheme to use 0.1.1 and 0.1.2 which will work for sure

lewisdiamond: `-` is for pre-release, it will be ignored by anything that excludes pre-release. We're putting the git commit id of the repo, so `+` (build metadata) seems more appropriate.

lewisdiamond: We should be using 0.1.1, 0.1.2 as release scheme, disjoint from the appVersion. This seems to be the standard anyways.

TerryHowe (approved):
/lgtm

PR #604, opened 2022-03-24
Add source registry

TerryHowe:
Use the sourceRegistry for the test app


lewisdiamond:
Why do we want to rename that to `sourceRegistry` and move it to the top level?

PR #617, opened 2022-03-29
Use mac as dhcp-identifier in netplan config file 

yxinchen:
*Issue #, if available:*

*Description of changes:*
Use mac as dhcp-identifier in netplan config file, this is to prevent ip change after instance reboot on Snow.

*Testing:*
Tested with beta device, confirmed instance ip persisted after instance stop/start.


g-gaston (approved):
/lgtm

PR #629, opened 2022-03-31
Install reboot in tink-docker

pokearu:
*Description of changes:*
tink-docker has a reboot watch, that calls `/sbin/reboot`.  `reboot` is a symlnk to `systemctl` and its dependencies.
Hence installing `systemd` fully on tink-docker.


abhay-krishna (approved):
/lgtm

PR #631, opened 2022-03-31
Fix the pipeline build for charts

TerryHowe:
Couple things going on here, build works fine locally, but failing in pipeline

* pipeline builds in private and pushes to public
* we want the helm chart to pull from public
* the images are not available in private, but they are promoted to public after the build
* since the images may be coming from private or public, we check both

### DIFF
--- a/build/lib/helm_require.sh
+++ b/build/lib/helm_require.sh
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-HELM_REGISTRY="${1?First argument is registry}"
+IMAGE_REGISTRY="${1?First argument is registry}"
 HELM_DESTINATION_REPOSITORY="${2?Second argument is helm destination repository}"
 OUTPUT_DIR="${3?Third argument is output directory}"
 IMAGE_TAG="${4?Fourth argument is image tag}"
@@ -45,7 +45,7 @@ spec:
 REQUIRES_CONFIG_FILE=helm/requires-config.yaml
 SEDFILE=${OUTPUT_DIR}/helm/sedfile
 export IMAGE_TAG
-export HELM_REGISTRY
+export HELM_REGISTRY=$(aws ecr-public describe-registries --region us-east-1  --output text --query 'registries[*].registryUri')
 envsubst <helm/sedfile.template >${SEDFILE}
 # Semver requires that our version begin with a digit, so strip the v.
 echo "s,version: v,version: ,g" >>${SEDFILE}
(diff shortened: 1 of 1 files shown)


build/lib/helm_require.sh
```diff
@@ -45,15 +45,22 @@ spec:
 REQUIRES_CONFIG_FILE=helm/requires-config.yaml
 SEDFILE=${OUTPUT_DIR}/helm/sedfile
 export IMAGE_TAG
-export HELM_REGISTRY
+export HELM_REGISTRY=$(aws ecr-public describe-registries --region us-east-1  --output text --query 'registries[*].registryUri')
```
ewollesen, line 48: Is it possible to have two registries? Maybe it's safer to select index 0? It might not be the one we want, but at least the script wouldn't barf.

Also, no check is done here for a failure in the aws command, or a blank return:
```bash
if [ -z "$HELM_REGISTRY" ]; then
    logger --stderr --tag "helm_require.sh" "no helm registry found via lookup"
fi
```

build/lib/helm_require.sh
```diff
@@ -45,15 +45,22 @@ spec:
...
 for IMAGE in ${HELM_IMAGE_LIST:-}
 do
-  IMAGE_SHASUM=$(${SCRIPT_ROOT}/image_shasum.sh ${HELM_REGISTRY} ${IMAGE} ${LATEST})
+  if [ "${IMAGE}" == "${HELM_DESTINATION_REPOSITORY}" ]
+  then
+    TAG="${IMAGE_TAG}"
+  else
+    TAG="${LATEST}"
+  fi
+  IMAGE_SHASUM=$(${SCRIPT_ROOT}/image_shasum.sh ${HELM_REGISTRY} ${IMAGE} ${TAG}) ||
```
ewollesen, line 62: Do you really want to push to helm registry, but not the image registry? That seems unusual.

ewollesen, line 62: Might want to only push if `$HELM_REGISTRY` isn't blank.

build/lib/helm_require.sh
ewollesen, line 48: It feels a little strange to me that we're specifying a registry, but not pushing to it if the push to a helm registry that configure via script succeeds. Is that really the intended behavior? How does the user know which registry received the push?

TerryHowe: At this time ECR public does not support more than one

TerryHowe: There are several steps in a helm build. The push comes later. All the steps are run with a `make helm/push`

build/lib/helm_require.sh
TerryHowe: The problem here is in the pipeline, the images are built to the private registry and *after* the build there is a step to promote them to the public registry. We need to get the latest images from private and have the helm chart pull them from public.

build/lib/helm_require.sh
TerryHowe: above there are a bunch of bash settings that are *supposed* to fail the entire script if a part fails.

build/lib/helm_require.sh
TerryHowe: The `image_shasum.sh` script will  fail with a parameter not set error

PR #632, opened 2022-03-31
Adding busybox reboot to tink-docker

pokearu:
*Description of changes:*
We need the reboot call from busybox binary instead of the systemctl. This enables reboots without an init daemon.


abhay-krishna (approved):
/lgtm

PR #635, opened 2022-04-01
govc: update to 0.27.4 and add validation for cli version

jaxesn:
*Issue #, if available:*

*Description of changes:*

Review by commit:

- While running through and updating the readme I decided to add to add the docker helpers to the common.mk to avoid having to run from the root of the repo.  I also updated the help.mk files for this change.
- Updates govc to 0.27.4. I did not see any breaking issues in the changelogs that seem like they would affect the way we are using the cli.  The version of go was bumped to 1.17.  Upstream now uses ldflags to set the version/commit/build date, I have added the validation targets we recently added to eks-d to make sure we continue to set these properly.


abhinavmpandey08 (approved):
/lgtm

PR #638, opened 2022-04-01
Adding CAPT patch to update PBNJ version

pokearu:
*Description of changes:*
CAPT was using an older version of PBNJ that had a different proto API.
Update fixes issues with wrong API calls being triggered.


abhinavmpandey08 (approved):
/lgtm

PR #644, opened 2022-04-02
kind: upgrade to 0.12.0

jaxesn:
*Issue #, if available:*

*Description of changes:*

This updates kind to 0.12.0 from 0.11.1.  Added the updating section to the rebuild and updated the various constants which are replaced during the build to override image tags.  Bumped golang to 1.17.  Removes iptables alternatives call since this image is now based off the iptables minimal image which has this same code.

This improves the way images are loaded by removing the kind code that uses ctr pull for non-kube images and instead loads them all after the fact as import so unpacking can be skipped.  For some reason this was causing issues with the 1.22 images, I *think* because the pause image in 1.22 is not based on scratch so the fact that it was being loaded with a snapshot and we weren't preserving these snapshots was causing issues when trying to launch pods.  I can not explain why this just popped up with this new version of kind, tho I suspect its some kind of bug in containerd regarding the sandbox image and the new fuse overlay plugin kind is configuring.


TerryHowe (approved):
/lgtm

TerryHowe (approved):
/lgtm

PR #649, opened 2022-04-03
update go install script based on builder base versions in periodic

jaxesn:
*Issue #, if available:*

*Description of changes:*

This will help catch missing projects from the ALL_PROEJCTS list and otherwise broken project builds, esp in release branches, by automatically updating the install go script, which is really just a helper for local machine set, but updating this file will trigger all presubmit jobs.  Whenever go versions are bumped in builder base, the next attribution job will include updates to this script in the checksums PR.

/hold


build/lib/update_go_versions.sh
```diff
@@ -0,0 +1,33 @@
...
+
+GOPATH=$GOPATH
+if [ -d "/go" ]; then
+    GOPATH="/go"
+fi
+
+for go in $GOPATH/go*/bin/go; do
+    VERSION=$($go version | grep -o "go[0-9].* " | grep -o "[0-9\.]*")
+    NOPATCH=$(cut -d. -f"1,2" <<< $VERSION)
+    sed -i "s,\${GOLANG${NOPATCH//./}_VERSION:.*},\${GOLANG${NOPATCH//./}_VERSION:-$VERSION}," ./build/lib/install_go_versions.sh
```
abhay-krishna, line 32: I suppose it shouldn't matter but do we need a `-` here?
```suggestion
    sed -i "s,\${GOLANG${NOPATCH//./}_VERSION:-.*},\${GOLANG${NOPATCH//./}_VERSION:-$VERSION}," ./build/lib/install_go_versions.sh 
```

jaxesn: It shouldnt matter, but i dont mind having it.

PR #656, opened 2022-04-04
Implement retry logic for tink-worker docker image pull...

ptrivedi:
...to allow for races where linuxkit network or dns may not have been fully set up and functional yet.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/projects/tinkerbell/hook/patches/0002-Implement-retry-for-tink-worker-docker-image-pull-to.patch
+++ b/projects/tinkerbell/hook/patches/0002-Implement-retry-for-tink-worker-docker-image-pull-to.patch
@@ -1,0 +1,47 @@
+From f5130587c14ba6f319c1435a132f0ca31d8aad0a Mon Sep 17 00:00:00 2001
+From: Pooja Trivedi <poojatrivedi@gmail.com>
+Date: Mon, 4 Apr 2022 22:21:23 +0000
+Subject: [PATCH] Implement retry for tink-worker docker image pull to allow
+ for races where linuxkit network or dns may not have been fully set up and
+ functional yet.
+
+---
+ bootkit/main.go | 22 +++++++++++++++++++---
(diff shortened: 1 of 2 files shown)


projects/tinkerbell/hook/patches/0002-Implement-retry-for-tink-worker-docker-image-pull-to.patch
```diff
@@ -0,0 +1,47 @@
...
++++ b/bootkit/main.go
+@@ -137,9 +137,25 @@ func main() {
+ 
+ 	fmt.Printf("Pulling image [%s]", imageName)
+ 
+-	out, err := cli.ImagePull(ctx, imageName, pullOpts)
+-	if err != nil {
+-		panic(err)
++	// TODO: should remove all this hardcoding or move it to a const at some point
++	attempts := 10
```
jacobweinstock, line 24: in my opinion, bootkit in general should behavior more like a kubernetes controller. Try to do all the things it currently does, but on failure just try again. and again, forever.  This is most certainly a longer term thing to think about but wanted to bring it regardless.

This seems like a very reasonable step forward, in general.

jacobweinstock (approved):
lgtm

projects/tinkerbell/hook/patches/0002-Implement-retry-for-tink-worker-docker-image-pull-to.patch
ptrivedi: > in my opinion, bootkit in general should behavior more like a kubernetes controller. Try to do all the things it currently does, but on failure just try again. and again, forever. This is most certainly a longer term thing to think about but wanted to bring it regardless.
> 
> This seems like a very reasonable step forward, in general.

Agree, we did debate about having it try forever or not. Thanks.

abhinavmpandey08 (approved):
/lgtm

PR #665, opened 2022-04-07
Upgrade helm version

a-cool-train:
*Issue #, if available:*

*Description of changes:*


jaxesn:
Can you add a readme to this project with an updating section? We have started to do this elsewhere so you can check one of the projects and use that updating section as a guide.  The biggest part I want to make sure is documented is the file upstream you are using as the golang version source of truth.

a-cool-train:
> Can you add a readme to this project with an updating section? We have started to do this elsewhere so you can check one of the projects and use that updating section as a guide. The biggest part I want to make sure is documented is the file upstream you are using as the golang version source of truth.

I follow the first part of the response but not the latter half. What do we want to document for golang version source of truth? As for the patch changes, I have add the source of the patch file in the patch itself.

a-cool-train:
> > Can you add a readme to this project with an updating section? We have started to do this elsewhere so you can check one of the projects and use that updating section as a guide. The biggest part I want to make sure is documented is the file upstream you are using as the golang version source of truth.
> 
> I follow the first part of the response but not the latter half. What do we want to document for golang version source of truth? As for the patch changes, I have add the source of the patch file in the patch itself.

Ignore this comment.

jaxesn (approved):
/lgtm

PR #680, opened 2022-04-11
Put packer config files at the appropriate location where packer builds look for them

abhay-krishna:
Placing the packer config files at the appropriate location where packer builds look for them. Fixes image builds looking for missing files.


pokearu (approved):
/lgtm

PR #681, opened 2022-04-12
Fix packer log path for OVA builds

abhay-krishna:
This puts the packer log in a subfolder based on the OS - Ubuntu, RHEL, etc.


pokearu (approved):
/lgtm
/woof

PR #700, opened 2022-04-14
Add RHEL 8 to RHEL versions list in image-builder Makefile

abhay-krishna:
We need to add rhel-8 to RHEL_VERSIONS list to be able to call the new RHEL 8 ova target.


vignesh-goutham (approved):
oh good find

/lgtm

PR #708, opened 2022-04-16
avoid assign shell calls in makefile macros

jaxesn:
*Issue #, if available:*

*Description of changes:*
`:=` in makefiles creates a "real" variable who's value is determine at the time of assignment.  `=` or `?=` creates a macro in the makfile who's value is not determined until the time that it is to be used.  We use this intentionally all over to ensure we can set variables during the flow through make targets.  When calling `$(shell...)` via a macro, make shells out everytime the value is requested.  For something like `BASE_DIRECTORY` or `GIT_TAG` this could be a number of times.  This changes as many of the calls to shell as possible to use `:=` instead of `=`.  There are still some instances throughout the code base of `=$(shell ..)` which are completely acceptable and necessary.  For example, any time the shell needs the value of a var which may not have been set yet or needs to access the REPO which hasnt been cloned yet.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -6,7 +6,7 @@ SHELL=bash
 
 RELEASE_ENVIRONMENT?=development
 
-GIT_HASH=$(shell git -C $(BASE_DIRECTORY) rev-parse HEAD)
+GIT_HASH:=$(shell git -C $(BASE_DIRECTORY) rev-parse HEAD)
 
 COMPONENT?=$(REPO_OWNER)/$(REPO)
 MAKE_ROOT=$(BASE_DIRECTORY)/projects/$(COMPONENT)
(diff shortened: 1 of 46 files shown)


chrisdoherty4:
Are we sure we want to swap out many of the `?=` assignments? `?=` is useful for parameterizing invocations of make recipes while supplying sane defaults. E.g. `make build GIT_TAG="custom-tag-for-hack-build"`.

Disclaimer: I don't know the (original) intent for many of these Make variables. Perhaps uses of `?=` were accidental.

jaxesn:
We def will continue to use ?= since we use it all over the place in Common.mk intentionally for its set if not set properties.  This specific usage is mostly likely a hold over from the earliest days of the eks-d makefiles where I do not believe we had specific intention one way or another and were just using ?= everywhere.  

That said, your example actually still works as intended without the `?=`.  The `?=` is more about catching values from Environment vars or in cases where the makefile may have set it previously.  Vars passed to make directly will override variables being set (unless maybe if you use the override directive?).

`make build GIT_TAG="custom-tag-for-hack-build"` - GIT_TAG from cli arg takes priority
`GIT_TAG="custom-tag-for-hack-build" make build` or `export GIT_TAG="custom-tag-for-hack-build" && make build` - GIT_TAG is set to the value from the file.  If the assignment was `?=` this syntax would also take priority.

PR #718, opened 2022-04-21
Use HTTP server as boot media path for RHEL 8 OVA builds

abhay-krishna:
RHEL 8 removed floppy support for automated installation, so we have to use the local HTTP server provisioned by Packer to serve the kickstart config for OS installs.


abhay-krishna:
/unhold

Tested these changes on Codebuild and it works

vignesh-goutham (approved):

PR #721, opened 2022-04-26
flux: bump to latest versions

jaxesn:
*Issue #, if available:*

*Description of changes:*
Bumping flux and its controllers to the latest versions.  This required a few extra changes

- helm/source-controller dep on the distribution/distrubition@2.8.0 which has a broken checksum from upstream. added patches to replace with 2.8.1 which is the same content just fixed checksum
- kustomize-controller showed an issue with the way we are calling go-licenses due to a change in syntax between 1.16 and 1.17 which requires that we use 1.17 for go-licenses.  https://github.com/aws/eks-distro-build-tooling/pull/378 adds go-licenses for 1.17 and this PR introduces handling for these cases
- source-controller upstream bumped the version of libgit2 to 1.3.1. https://github.com/aws/eks-distro-build-tooling/pull/379 bumps this in our base image


projects/fluxcd/kustomize-controller/GIT_TAG
```diff
@@ -1 +1 @@
-v0.19.0
\ No newline at end of file
+v0.24.1
```
jiayiwang7, line 1: shouldn't this be `v0.24.3`?

jiayiwang7 (commented):
Thanks!

jaxesn:
eks-anywhere-cli-tools-tooling-presubmit will fail until we actually merge the flux bump

/override eks-anywhere-cli-tools-tooling-presubmit

EKS_DISTRO_MINIMAL_BASE_GIT_TAG_FILE
```diff
@@ -1 +1 @@
-2022-04-11-1649641850.2
+2022-04-27-1651020237.2
```
jaxesn, line 1: I am checking this in along with these changes because the new version of libgit2 is only compatible with this new version of the source controller.

abhay-krishna (approved):
/lgtm

PR #724, opened 2022-04-27
Increase file monitoring timeout to 40s

johnsonjnpan:
*Description of changes:*
Currently, the file monitoring method timeout after 10 seconds, this timeout did not work for our use-case, due to the fact that we rely on `/etc/hosts` to point the cluster control plane endpoint to their IPs (so that we support offline operation)
E.g.
```
10.0.0.3 cluster.domain.local
10.0.0.2 cluster.domain.local
10.0.0.1 cluster.domain.local
```
During joining of the etcd, `kubeadm` will try to connect to each IP in the above order, until it successfully establish a TCP connection. In case where only the last IP (10.0.0.1) has the valid APIServer up, it will take `kubeadm` quite some time (> 10s) to eventually try to connect to that IP.
Based on what we try, we found that setting the timeout to a higher value (e.g. 40s) solves our issue.
Request bumping the timeout to 40s.


jaxesn:
this go-licenses error is my fault. i am going to revert the prow job changes soon and this should pass on retest. sorry for the churn

vignesh-goutham:
> that going to cover the vast majority of cases or are you still

As I understand it, this is only to cover one of the corner cases. The 10 second works for most of the cases. @johnsonjnpan showed me an instance where there were 3 CPI and if the bootstrapped CPI was the last on the /etc/host list its takes sometime for kubeadm to hit each endpoint to get the joinConfig. 

@johnsonjnpan the time required would go up when there are more CPIs right or is the CPI number there static?

johnsonjnpan:
> @johnsonjnpan the time required would go up when there are more CPIs right or is the CPI number there static?

Theoretically speaking, if we add more CPIs, let say 5, which means we will also have 5 `ip:host` mappings in `/etc/hosts`. Then yes, it would take more time for the OS find out the valid IP to connect.

Based on my experience using `netcat`, it takes ~5 seconds per each invalid IPs that it try to connect.

Currently we always create clusters with 3 CPIs, so this timeout should be sufficient for us.

PR #727, opened 2022-04-28
Fix source controller

jaxesn:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/build/lib/buildkit.sh
+++ b/build/lib/buildkit.sh
@@ -47,4 +47,3 @@ else
     # skip retry when running locally
     $CMD "$@"
 fi
-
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

hard to imagine how this could happen

jaxesn:
I see what happened. The builder base built twice today, once this morning after fixing the go-licenses issue. This included gcc package `7.3.1-13.amzn2`.  The second time it built this afternoon was when buildkit was updated.  This build also happened to get a newer gcc `7.3.1-14.amzn2`.  This isnt an issue at all, however, since source-controller is a cgo binary the gcc version ends up getting embedded in the final binary (i verified this running `strings` against the binary).  This is what is causing the checksum change.  I have confirmed these new checksums should be consistent.  I am going to investigate stripping that version info out of the final binary to avoid this in the future.

PR #754, opened 2022-05-03
Generate tink manifest

abhinavmpandey08:
*Description of changes:*
This PR adds the `create-manifests` target to the `tinkerbell/tink` project Makefile which then generates the Kubernetes manifests for `tink-controller` and `tink-server`.
Also, removes the `.ONESHELL` dependency in the upstream Makefile 


projects/tinkerbell/tink/build/create_manifests.sh
```diff
@@ -0,0 +1,44 @@
...
+GOLANG_VERSION="$7"
+
+MAKE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+source "${MAKE_ROOT}/../../../build/lib/common.sh"
+
+build::common::use_go_version ${GOLANG_VERSION}
+
+cd $REPO
+
+npm install prettier --global
```
jaxesn, line 36: What are they using this for?

projects/tinkerbell/tink/patches/0001-Remove-.ONESHELL-from-rules.mk.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/rules.mk b/rules.mk
+index a881c42..027f521 100644
+--- a/rules.mk
++++ b/rules.mk
+@@ -5,8 +5,6 @@ MAKEFLAGS += --no-builtin-rules
+ # This avoids a failure to then skip building on next run if the output is created by shell redirection for example
+ # Not really necessary for now, but just good to have already if it becomes necessary later.
+ .DELETE_ON_ERROR:
+-# Treat the whole recipe as a one shell script/invocation instead of one-per-line
+-.ONESHELL:
```
jaxesn, line 20: What issue did this cause?

projects/tinkerbell/tink/build/create_manifests.sh
abhinavmpandey08: `prettier` is used at a bunch of places in tink's mk to beautify the generated YAMLs and stuff https://github.com/tinkerbell/tink/blob/main/kube.mk#L63
I don't think it is required or anything since all of these manifests are auto-generated using controller-gen and kustomize so I could patch that out if needed.

projects/tinkerbell/tink/patches/0001-Remove-.ONESHELL-from-rules.mk.patch
abhinavmpandey08: Apparently the make version in builder-base has a bug where `.ONESHELL` and `. SHELLFLAGS` don't play well with each other. The error is something like `invalid option name`

Here's the stackoverflow question for reference https://stackoverflow.com/questions/68040084/oneshell-in-makefile-not-working-with-bash-on-amazon-linux-2-ami/68041602#68041602

projects/tinkerbell/tink/build/create_manifests.sh
jaxesn: no i dont think it matters, just dont usually see go projects pulling in anything but other go utils, esp not a npm package!

projects/tinkerbell/tink/patches/0001-Remove-.ONESHELL-from-rules.mk.patch
jaxesn: oh wow, interesting, kind of surprised we havent seen this before.  Well switch to an al22 base in the summer after they ga which has a newer make version.

abhinavmpandey08: @pokearu mentioned he had seen this issue before when trying to build boots but yeah newer make version should fix this. Tried with `GNU Make 4.2.1` and no issues

projects/tinkerbell/tink/build/create_manifests.sh
abhinavmpandey08: yeah it is weird, prettier only has npm and yarn packages

jaxesn (commented):
this looks good to me

projects/tinkerbell/tink/patches/0001-Remove-.ONESHELL-from-rules.mk.patch
abhay-krishna: Tbh even an older make version would fix it I think. Make v3.82 is just terrible but AL2 wouldn't update

PR #762, opened 2022-05-05
Build eks-anywhere-packages and include it in releases

ewollesen:
Trying to get eks anywhere packages included in the build/release process.

### DIFF
--- a/release/staging-build.yml
+++ b/release/staging-build.yml
@@ -205,3 +205,9 @@ batch:
     depend-on:
     - cri_tools
     - etcdadm
+  - identifier: eks-anywhere-packages
+    buildspec: buildspec.yml
+    env:
+      variables:
+        PROJECT_PATH: projects/aws/eks-anywhere-packages
+        CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/aws.eks-anywhere-packages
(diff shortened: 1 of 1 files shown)


release/staging-build.yml
```diff
@@ -205,3 +205,9 @@ batch:
     depend-on:
     - cri_tools
     - etcdadm
+  - identifier: eks-anywhere-packages
```
abhay-krishna, line 208: Identifier names cannot have dashes.
```suggestion
  - identifier: eks_anywhere_packages
```

(I know right 🙄)

release/staging-build.yml
```diff
@@ -205,3 +205,21 @@ batch:
     depend-on:
     - cri_tools
     - etcdadm
+  - identifier: eks_anywhere_packages
```
abhay-krishna, line 208: let's reorder to put this at the bottom, and add a depend-on field to the packages build
```yaml
depend-on:
- harbor
- hello_eks_anywhere
```

line 205 has an example of adding such a clause, and its indentation.

ewollesen: Oh shoot, right, obviously I forgot that part. Thank you.

abhay-krishna: this is because identifiers used in the `depend-on` section need to be defined before being used.

ewollesen: Heh, I was kinda wondering if there were ordering requirements.

abhay-krishna (commented):
/lgtm
/approve

PR #767, opened 2022-05-06
Add target to build latest snow raw admin image

g-gaston:
Part of https://github.com/aws/eks-anywhere/issues/1166

## Description of changes
This image will be built from a pipeline. Such pipeline will run whenever the release manifest file in S3 changes (this happens every time we build a new version). But that doesn't tell us which is the version that just got built.

We could use the `latestVersion` in the manifest but that will be pointing to the latest release for the latest minor version. If we build a new patch release for the previous minor version (eg. `latestVersion` is `v0.8.2` and we build `v0.7.5` due to some vulnerability) then the `latestVersion` field won't change.

Instead, this just iterates over all the releases and picks the latest one, by date.

This is not ideal. If we trigger two releases in parallel very close in time, it's possible that we will build only one image for whichever release finishes last. We haven't done this before and probably ever won't, so this mechanism should work for now. But something to take into consideration in the future.

### DIFF
--- a/projects/aws/eks-a-admin-image/build/get_latest_eksa_version_by_date.sh
+++ b/projects/aws/eks-a-admin-image/build/get_latest_eksa_version_by_date.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
+RELEASES_MANIFEST=$(curl --silent -L $EKSA_RELEASE_MANIFEST_URL)
+
+newest_release_date=0
+newest_release_version=""
+
+while IFS=$'\t' read -r date version _; do
+	# date is something like '2022-05-05 13:05:34.038243612 +0000 UTC'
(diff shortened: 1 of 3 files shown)


projects/aws/eks-a-admin-image/build/get_latest_eksa_version_by_date.sh
```diff
@@ -0,0 +1,30 @@
...
+
+while IFS=$'\t' read -r date version _; do
+	# date is something like '2022-05-05 13:05:34.038243612 +0000 UTC'
+	# I can't get `date` to parse this format, so deleting the extra timezone first
+	parsed_date=$(date -d "$(echo $date | awk '{print $1,$2,$3}')" +"%s")
+
+	if [ $parsed_date -gt $newest_release_date ];
+	then
+	    newest_release_date=$parsed_date
+		newest_release_version=$version
```
jaxesn, line 21: Couple indention issues here?

jaxesn (approved):

PR #769, opened 2022-05-06
Adds new target to create ecr repos

jaxesn:
*Description of changes:*
When building locally it can be useful to push built images to ecr
in our aws accounts to make testing easier. create-ecr-repos target
will create repos for each image defined in the project's IMAGE_NAMES.
This also adds a check during buildctl calls running locally for a
common EOF error that typically indicates a missing repo.

### DIFF
--- a/build/lib/buildkit.sh
+++ b/build/lib/buildkit.sh
@@ -45,5 +45,15 @@ if [ -f "/buildkit.sh" ]; then
     (exit $s)
 else
     # skip retry when running locally
-    $CMD "$@"
+    log_file=$(mktemp)
+    trap "rm -f $log_file" EXIT
+    if ! $CMD "$@" 2>&1 | tee $log_file; then        
+        if grep -q "blobs/uploads/\": EOF" $log_file ; then
+            echo "******************************************************"
+            echo "Ensure container registry and repository exists!!"
+            echo "Try running make create-ecr-repos to create ecr repositories in your aws account."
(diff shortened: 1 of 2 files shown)


build/lib/buildkit.sh
```diff
@@ -45,5 +45,15 @@ if [ -f "/buildkit.sh" ]; then
     (exit $s)
 else
     # skip retry when running locally
-    $CMD "$@"
+    log_file=$(mktemp);
```
ewollesen, line 48:
```suggestion
    log_file=$(mktemp);
    trap "rm -f $log_file" EXIT
```
Is it worth cleaning up after ourselves?

build/lib/buildkit.sh
```diff
@@ -45,5 +45,15 @@ if [ -f "/buildkit.sh" ]; then
...
+    if ! $CMD "$@" 2>&1 | tee $log_file; then        
+        if grep "blobs/uploads/\": EOF" $log_file > /dev/null 2>&1; then
```
ewollesen, line 50:
```suggestion
        if grep -q "blobs/uploads/\": EOF" $log_file ; then
```
Just a little cleaner.

ewollesen (commented):
Just a couple of comments. I could /lgtm as is if you don't like the suggestions.

TerryHowe:
Nice, I need to remember this exists.

PR #772, opened 2022-05-06
Select Availability zone randomly for EC2 instance creation

abhay-krishna:
It always chooses us-west-2b, which causes InsufficientCapacity issue as below
```
An error occurred (InsufficientInstanceCapacity) when calling the RunInstances operation (reached max retries: 2): We currently do not have sufficient i3en.metal capacity in the Availability Zone you requested (us-west-2b). Our system will be working on provisioning additional capacity. You can currently get i3en.metal capacity by not specifying an Availability Zone in your request or choosing us-west-2a, us-west-2c, us-west-2d.
```

### DIFF
--- a/projects/kubernetes-sigs/image-builder/build/build_raw_image.sh
+++ b/projects/kubernetes-sigs/image-builder/build/build_raw_image.sh
@@ -81,9 +81,14 @@ MAX_RETRIES=15
 for i in $(seq 1 $MAX_RETRIES); do
     echo "Attempt $(($i))"
 
+    AZ_LIST="us-west-2a,us-west-2b,us-west-2c,us-west-2d"
+    AZ_COUNT=$(echo $AZ_LIST | awk -F\, '{print NF}')
+    INDEX=$((($RANDOM % $AZ_COUNT) + 1))
+    AZ=$(cut -d',' -f${INDEX} <<< $AZ_LIST)
+
     # Create a single EC2 instance with provided instance type and AMI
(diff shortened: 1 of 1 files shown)


projects/kubernetes-sigs/image-builder/build/build_raw_image.sh
```diff
@@ -81,9 +81,15 @@ MAX_RETRIES=15
 for i in $(seq 1 $MAX_RETRIES); do
     echo "Attempt $(($i))"
 
+    AZ_LIST=($(aws ec2 describe-availability-zones --query "AvailabilityZones[*].ZoneName" --output text))
```
jaxesn, line 84: didnt we have this problem before somewhere else?

abhay-krishna: i think kops tests?

jaxesn: hmm no, im thinking maybe in the e2e tests framework? idk tho, but i feel like weve had this discussion before....

abhay-krishna: I think that was throttling based on requests to the RunInstances API call, which we fixed by implementing retries. I feel like this is the first time I've seen the insufficient capacity one, but I may be mistaken.

jaxesn (approved):
Let's give a shot!

/approve

PR #788, opened 2022-05-10
add ssm activation to eks-a-admin ci image

gwesterfieldjr:
Adding script to activate ssm on boot. During OVF deployment in vSphere I inject ssm_actication_code, ssm_actication_id, and ssm_actication_region as ovf properties to the VM. This script picks them up and activates the node.

PR #794, opened 2022-05-11
Create helm chart repos

ewollesen:
The create-ecr-repos Makefile target is extended to create the repos
of helm charts as well. In many cases the helm chart will end up in
one of the repos that's already created for a container image, but in
other cases, they're different repos, and this code will handle that
situation.

*Issue #, if available:*

*Description of changes:*


jaxesn (approved):
I like this!

TerryHowe:
failing test looks like

gwesterfieldjr:
https://github.com/aws/eks-anywhere-build-tooling/pull/788/commits/253fd6f3a11a5d5ceed4d371222bf3b86b5381ed

Apply these changes to fix presubmit. Looks like packer validate does not like default of null values.

PR #817, opened 2022-05-17
Skip checksum validation for bootconfig

vignesh-goutham:
*Description of changes:*
Bootconfig is built using gcc. The current version of gcc used on builder-base does not support flags to do relative directory builds, hence resulting in non-deterministic builds. This change is to skip the checksum validation till the gcc gets updated.


jaxesn (approved):
/approve

PR #820, opened 2022-05-17
Remove Bootconfig tarball prefix

vignesh-goutham:


jaxesn:
You sure you want to do this? The resulting file name will be kind of wierd without it

PR #823, opened 2022-05-18
Supprt custom tar file prefixes when fetching dep binaries

jaxesn:
*Issue #, if available:*

*Description of changes:*

For some projects, like bootconfig, we use a custom name for the tarball. Currently the fetch binaries helper script is unaware of this support and assumes it named a certain way.  This adds support for fetching these custom names tarballs.

In cases where we are not building arm versions, which should be avoided but is currently the case for bootconfig, we need to not try and pull arm tars.  This changes that to use the BINARY_PLATFORMS var as the source of truth for which tars to pull.  Instead we may want/need to introduce a FETCH_BINARIES_PLATFORMS, but id rather avoid it if we dont need it.


vignesh-goutham (approved):
/approve

PR #830, opened 2022-05-18
Package BootConfig with Writefile action

vignesh-goutham:
*Description of changes:*
Along with packaging bootconfig with writefile action, this PR cleans up `IMAGE_PLATFORMS`. With this we can override the Image platforms for which we build and push.

I'm not entirely sure if we should merge both `BINARY_PLATFORMS` and `IMAGE_PLATFORMS`, although I cant think of a case where we'll require variables for each in a particular build. 

/hold


Common.mk
```diff
@@ -142,6 +142,7 @@ IMAGE_USERADD_USER_NAME?=
 IMAGE_IMPORT_CACHE?=type=registry,ref=$(LATEST_IMAGE) type=registry,ref=$(subst $(LATEST),latest,$(LATEST_IMAGE))
```
jaxesn, line 142: I think these changes are reasonable and I would like to clean up the image_platform stuff, but I'm not actually sure you need these changes at all?  Overriding IMAGE_PLATFORMS in your project MAKEFILE should take priority over the values in each of these targets.  As far as the IMAGE_TARGETS_FOR_NAME, having the extra arm64 in there should hurt anything since that is just a helper to make it easier to add pre-reqs to image builds.  Can you confirm this one way or another?

As an aside, i know we dont need it and arent testing it yet, but beyond that obvs reason, any reason we couldnt just build an arm binary of bootconfig and avoid having to skip the arm image builds altogether?  Def could be a follow discussion

jaxesn (approved):

PR #833, opened 2022-05-19
retries docker pull and use builder base tag

jaxesn:
Mitali ran into a couple issues during the staging build.

We got hit by the random tls handshake timeout when trying to pull
the qemu image and since the builder base had been updated with new
golang versions, the source controller build pulled the latest
builder base and the checksums did not match. Abhay had added the
builder base tag env var and config file, for code build and prow
respectively, but we never implemented using them. With this change
the source controller cgo build should always pull the same builder
base that the current build is running in, instead of the latest.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ stop-docker-builder:
 
 .PHONY: run-buildkit-and-registry
 run-buildkit-and-registry:
-	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.9.0-rootless
+	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.10.3-rootless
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -94,7 +94,7 @@ stop-docker-builder:
 
 .PHONY: run-buildkit-and-registry
 run-buildkit-and-registry:
-	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.9.0-rootless
+	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.10.3-rootless
```
jaxesn, line 97: weve have been using 0.10 in the builder base, just updating this for local building.

Issue #801, opened 2022-05-13
Feature: Explicit dependency management


ewollesen:
At present, dependency management between projects is limited and implicit.

As an example, the harbor project is dependent on the distribution project. How is that dependency specified? It's a combination of non-obvious Makefile magic: [Loc #1](https://github.com/aws/eks-anywhere-build-tooling/blob/f9a041cdc13490f71fd59aebcc265b29969323e2/projects/goharbor/harbor/Makefile#L52) and [Loc #2](https://github.com/aws/eks-anywhere-build-tooling/blob/f9a041cdc13490f71fd59aebcc265b29969323e2/projects/goharbor/harbor/Makefile#L118-L119) and [Loc #3](https://github.com/aws/eks-anywhere-build-tooling/blob/f9a041cdc13490f71fd59aebcc265b29969323e2/Makefile#L8).

There is, at present, no way to specify any sort of version for the dependency. This means that Loc 3 above is ordering-dependent. If distribution isn't built before harbor, then harbor could be built with the incorrect version of distribution. That requirement isn't explicitly captured anywhere.

If the two projects are built in parallel, then the dependency is again possibly broken (e.g. `make -j$(nproc) build-all-projects`).

While the current system is functional, it seems very unlikely that it will be able to meet the future needs of the EKS-A Curated Packages team. Having the means to specify dependencies in an explicit manner, including versions or build identifiers will be important.

As an added bonus, we may be able to safely parallelize builds for a potential speedup.

/cc @jaxesn @vincentni

PR #841, opened 2022-05-20
Generate staging buildspec from Makefiles

jaxesn:
*Issue #, if available:*

*Description of changes:*

The main purpose of this PR is to provide explicit dependency declaration to support auto generating the staging-bundle.yml.

The staging build, which uses that bundle, is a batch codebuild which builds "everything" in the build-tooling repo.  The purpose of this is to have a process where everything is built from scratch, which is different than our dev build/release process which just builds the projects which changed. 

In some cases we have the build triggers, in our dev build process, setup where if dependent projects change, it will also trigger the project which depend on it, but I suspect this is in complete.  For the most this works out to be ok because we only have one version of the projects in this repo and dependencies are pulled using the GIT_TAG from repo.  That means that if say distribution changes to a new version and harbor kicks off before the distribution job publishes the new tarball, with the new git_tag in the name, it will [wait_for_tar](https://github.com/aws/eks-anywhere-build-tooling/blob/main/build/lib/fetch_binaries.sh#L67) to try and ensure it pulls the correct version. In the worst case this wait will timeout and the harbor build will fail, requiring a manual retrigger which would then use the newest version of the distribution dep.

The staging build/release building everything, using the batch build yml to model the dependencies, is to validate that everything is using the correct version and building in the proper order.  Currently the staging-build.yaml has been manually created and maintain and has gotten out of date, for example there is no distribution dependency on the harbor build as their should be.  The reason this ends up not being an issue is because the staging build builds to and pulls from the same folder/image tags as the dev build/release.  I want to explore changing this to ensure the staging build is running from a "clean" env.

Before introducing the staging-build.yaml we used to use https://github.com/aws/eks-anywhere-build-tooling/blob/main/Makefile#L8 to model these dependencies since we would build everything from one code build job, use `make release` from the main Makefile.  We do not do this anymore which makes this array of projects largely unnecessary and out of date.

We also had a psuedo-dependency model using the FETCH_BINARIES helper in makefiles which really just a way to standardize downloading the tars of other projects, there are really just a few examples of this across the repo.

The first commit in this PR changes FETCH_BINARIES into PROJECT_DEPENDENCIES with the main purpose of being able to use this information to regenerate the staging-bundle.yml ensuring it is always up to date and there is only one real source of truth for which projects depend on other projects.

*This has been removed for now, but I may bring it back in another PR* The second commit is more of a POC in how with deps modeled in the makefile structure, we could introduce an additional target to make it easier to build all the dependencies project locally if that is desired.

re: https://github.com/aws/eks-anywhere-build-tooling/issues/801

/hold


jaxesn:
/hold 

Adding presubmit to verify this stays up to date

build/lib/all_projects.sh
```diff
@@ -0,0 +1,39 @@
...
+cd $MAKE_ROOT
+
+# Iterating over orgs under projects folder
+repos=()
+for org_path in projects/*; do
+    org=$(cut -d/ -f2 <<< $org_path)
+    for repo_path in projects/$org/*; do
+        repo="$(cut -d/ -f3 <<< $repo_path)"
+        # couple odd ball non-builds in this repo
+        if [[ $repo = sandbox* ]]; then
```
abhay-krishna, line 32: possibly add an empty marker file in the oddball folders called `DO-NOT-BUILD` or something, that we check for here?

build/lib/generate_projects_list.sh
```diff
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
abhay-krishna, line 2: My bad, missed this

projects/aws/eks-a-admin-image/Makefile
```diff
@@ -45,7 +46,7 @@ PACKER=$(BINARY_DEPS_DIR)/packer
 build-%: export PKR_VAR_eks-a-version=$(EKSA_VERSION)
 build-%: export PKR_VAR_eks-a-release-manifest-url=$(EKSA_RELEASE_MANIFEST_URL)
 build-%: export PKR_VAR_build-version=$(shell echo $(GIT_HASH) | cut -c1-7)
-build-%: KUBECTL_URL:=$(shell ./build/get_kubectl_url.sh)
+build-%: KUBECTL_UR?=$(shell ./build/get_kubectl_url.sh)
```
abhay-krishna, line 49:
```suggestion
build-%: KUBECTL_URL?=$(shell ./build/get_kubectl_url.sh)
```

build/lib/generate_staging_buildspec.sh
```diff
@@ -0,0 +1,103 @@
...
+        fi
+        DEPEND_ON+="\"${DEP_ORG//-/_}_${DEP_REPO//-/_}\","
+
+        if [ ! -d $MAKE_ROOT/projects/$DEP_ORG/$DEP_REPO ]; then
+            echo "Non-existent project dependency: $dep!!!"
+            exit 1
+        fi
+    done
+
+    if [ ! -z "$DEPEND_ON" ]; then
```
abhay-krishna, line 76: equivalent to `if [ -n "$DEPEND_ON" ]; then` ?

ewollesen (commented):
Shoot, these comments have been sitting here for a while, sorry.

jaxesn:
/unhold

Added the presubmit

PR #855, opened 2022-05-27
Read local apiserver health check endpoint from apiserver manifest

johnsonjnpan:
*Description of changes:*
* The current health check in [BR-Bootstrap](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go#L77-L80) always checks the localhost’s 6443 port for the APIServer. However, we will be hosting our APIServer at port 443 instead. In this case, the health check will fail for our use case.
* This PR removes the hard coded port `6443` for health check. Instead it tries to parse the port number from the APIServer's static pod manifest
 * If it fails during parsing, it will fall back to port `6443` 


projects/aws/bottlerocket-bootstrap/pkg/kubeadm/utils.go
```diff
@@ -104,6 +104,31 @@ func getBootstrapFromJoinConfig(path string) (string, string, error) {
 	return "https://" + serverEndpoint, token, nil
 }
 
+func getLocalApiServerReadinessEndpoint() (string, error) {
+	data, err := ioutil.ReadFile("/.bottlerocket/rootfs/etc/kubernetes/manifests/kube-apiserver")
+	if err != nil {
+		return "", errors.Wrap(err, "Error reading ApiServer manifest file")
+	}
+
+	apiServerTree, err := unmarshalIntoMap([]byte(data))
```
vignesh-goutham, line 113: While this way is more resilient across different versions, we anyways use the v1 import for parsing Pod spec already. That would make this a bit easier to get to the scheme/host/port. What are your thoughts on using the v1.Pod struct here. An example [here](https://github.com/aws/eks-anywhere-build-tooling/blob/fcb1eaf1e2425cae4c1d73bd590e6a7f29b535b2/projects/aws/bottlerocket-bootstrap/pkg/utils/waiters.go#L178-L196)

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/utils.go
```diff
@@ -104,6 +104,31 @@ func getBootstrapFromJoinConfig(path string) (string, string, error) {
 	return "https://" + serverEndpoint, token, nil
 }
 
+func getLocalApiServerReadinessEndpoint() (string, error) {
+	data, err := ioutil.ReadFile("/.bottlerocket/rootfs/etc/kubernetes/manifests/kube-apiserver")
```
vignesh-goutham, line 108: Lets define this path as a constant.

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
```diff
@@ -73,11 +73,18 @@ func controlPlaneInit() error {
...
+		localApiServerReadinessEndpoint = "https://localhost:6443/healthz"
+	}
+
 	// Wait for Kubernetes API server to come up.
-	err = utils.WaitFor200("https://localhost:6443/healthz", 30*time.Second)
+	err = utils.WaitFor200(localApiServerReadinessEndpoint, 30*time.Second)
 	if err != nil {
 		return err
 	}
+	fmt.Printf("GotEndpoint: %s\n", localApiServerReadinessEndpoint)
```
vignesh-goutham, line 87: Probably from debugging? WaitFor200 anyways prints the url.

johnsonjnpan: Oops, removing.

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/utils.go
vignesh-goutham: There is a helper [here](https://github.com/aws/eks-anywhere-build-tooling/blob/fcb1eaf1e2425cae4c1d73bd590e6a7f29b535b2/projects/aws/bottlerocket-bootstrap/pkg/utils/parsers.go#L50) you can use

johnsonjnpan: Good idea, I also prefer using a proper type for this. 
Thanks for pointing that out!

vignesh-goutham (approved):
/approve

PR #876, opened 2022-06-03
Patch hook to skip pulling registry cert

abhinavmpandey08:
*Description of changes:*
Hook currently tries to pull the cert unconditionally and panics if the cert is not served from a hardcoded endpoint. 
This PR patches hook to only pull the cert if tls is enabled 


chrisdoherty4 (approved):
lgtm

PR #890, opened 2022-06-04
Add GIT_TAG for tinkerbell-helm

abhinavmpandey08:
*Description of changes:*
`GIT_TAG` is required for bundle release to work
Also adds some verification that runs helm lint and verifies the helm chart version matches the `GIT_TAG`


projects/tinkerbell/tinkerbell-chart/Makefile
```diff
@@ -17,8 +17,12 @@ RELEASE_TARGETS=publish-chart
 
 include $(BASE_DIRECTORY)/Common.mk
 
+.PHONY: verify
+verify:
+	scripts/verify.sh $(HELM_DIRECTORY) $(GIT_TAG)
+
 .PHONY: build-chart
-build-chart:
+build-chart: verify
```
jaxesn, line 25: Why do we have build-chart and publish-chart targets in this makefile? can we not use the ones from common.mk?   I prefer not having calls to the common scripts, like helm_copy/helm_build in these project makefiles, makes making changes to those harder later.

pokearu (approved):
/approve
/lgtm

PR #898, opened 2022-06-07
Update go-license installation instructions

TerryHowe:


jaxesn:
It may be worth calling this script this out? https://github.com/aws/eks-anywhere-build-tooling/blob/main/build/lib/install_go_versions.sh

I always use that to keep my golang versions the same on my local as in the builder-base, it also installs go-licenses.

docs/development/attribution-files.md
```diff
@@ -48,7 +48,7 @@ look out for.
 
 If trying to generate these files outside of CI and have them match what CI produces, you will need the following:
 
-1. `go get github.com/google/go-licenses`
+1. `go install github.com/google/go-licenses@latest`
```
TerryHowe, line 51:
```suggestion
1. `build/lib/install_go_versions.sh` will install the right go versions and go-licenses for that go version
```

PR #953, opened 2022-06-16
Fix README version field for projects with dashes in tag

abhay-krishna:
Some projects have a dash in their Git tag, and we can't use these directly in the badge URL template. Using two dashes renders the badge properly.
`https://img.shields.io/badge/version-v0.4.5-rc3-blue`: Badge
`https://img.shields.io/badge/version-v0.4.5--rc3-blue`: Badge


TerryHowe (approved):
/lgtm

PR #954, opened 2022-06-16
Security patch for Harbor

vincentni:
*Issue #, if available:*

*Description of changes:*


jaxesn (changes requested):
/hold

Could you document how you are producing these patches?

The attribution file does look correct, the go version specifically.  Its fine just to not include that change and let the periodic pick it up and update it for us.

jaxesn (approved):

Issue #959, opened 2022-06-17
Launch custom host-containers during bootstrap in CPIs


vpineda1996:
**What would you like to be added**:

I would like to add the ability to deploy host-level containers in my CPIs during the bootstrap process. Host-level containers are a feature inside of BR that allows us to launch "super-powered" containers inside the host. 

https://github.com/bottlerocket-os/bottlerocket#host-containers-settings

**Why is this needed**:

Currently there is no way to launch super-powered containers as regular pods/deployments/static pods are just regular containers that don't have root/superpower access.

PR #963, opened 2022-06-19
[release-0.9][Bugfix] Add bottlerocket control image on joining nodes

vpineda1996:
*Description of changes:*

Specifying bottlerocket control image was introduced by [this patch](https://github.com/aws/eks-anywhere-build-tooling/blob/release-0.8/projects/kubernetes-sigs/cluster-api/patches/0045-Support-configuring-bottlerocket-control-container-uri.patch). This feature is enabled for the first node but joining nodes will not respect the BR control image. My suspicion is that this bug was introduced after merge conflicts arose from merging upstream CAPI and this line was somehow omitted or lost.

This was discovered as I was working on #959 and noticed that it wasn't working as I expected it to.

#### Testing
I tried creating a new cluster with the patched version of BR and was successful at doing so.


vignesh-goutham (approved):
/approve

PR #990, opened 2022-06-21
Bump tink repository commit

chrisdoherty4:


UPSTREAM_PROJECTS.yaml
```diff
@@ -30,6 +30,8 @@ projects:
           - tag: v0.28.1
   - org: aws
     repos:
+      - name: cluster-api-provider-cloudstack
+        versions: []
```
ptrivedi, line 34: why is capc here??

pokearu:
We would need to update the [README#L2](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/tinkerbell/tink/README.md?plain=1#L2)

And the presubmit would fail and we need to update the CHECKSUMS file.

pokearu (approved):
/lgtm
/approve

PR #1000, opened 2022-06-22
Move kube-vip project to the kube-vip org

abhay-krishna:
Kube-vip moved to a new home at https://github.com/kube-vip, so adding a directory for running its builds. In a future PR, the `plunder-app` folder will be removed.


abhinavmpandey08 (approved):
/approve

PR #1027, opened 2022-06-30
Remove hardcoded region for SecretsManager

johnsonjnpan:
*Description of changes:*
The existing implementation of bottlerocket-bootstrap uses a hard coded AWS region string ("us-west-2") when accessing SecretsManager. This change removes the hardcoded string, instead, it loads it from the EC2 metadata.

This code path is ran only for remote user-data usecase.


projects/aws/bottlerocket-bootstrap/pkg/service/secretsmanager.go
```diff
@@ -17,12 +19,17 @@ type SecretsManagerImpl struct {
 	BaseClient *secretsmanager.Client
 }
```
TerryHowe, line 21: I'm not a big fan of the idea of a constructor returning an error, seems like an Init() method could have been added. Not a blocker for me though.

TerryHowe (commented):
/ok-to-test

projects/aws/bottlerocket-bootstrap/pkg/service/secretsmanager.go
johnsonjnpan: I just thought about the `init()` approach, it seems then we are just creating an `Init()` method and move most of the logic from `NewSecretsManagerService()` to the `Init()` method, which I not quite sure if it adds a lot of value.

Another approach will be, we change `NewSecretsManagerService()` into 
```
func NewSecretsManagerService2(cfg aws.Config, region string) (SecretsManagerService) {
	clientImpl := new(SecretsManagerImpl)
	cfg.Region = region
	clientImpl.BaseClient = secretsmanager.NewFromConfig(cfg)
	return clientImpl
}

```
But this will mean it'll be a bit verbose to use this class as we still need to construct `cfg` and `region` outside of this method.

What's your option?

vignesh-goutham: How about we default to a region when imds.GetRegion returns an error. I do not fully understand its implications, so feel free to dispute this. If this must return a proper region from imds then its fine I guess.

I like you above suggestion too, the config and region could be part of init(). But Im fine with what we have as well.

vignesh-goutham (commented):
LGTM. Please let me know if you are good to merge this.

johnsonjnpan:
Updated and ready to merge

vignesh-goutham (approved):
/approve

PR #1032, opened 2022-07-05
Add Cargo installation directory to PATH

abhay-krishna:
Codebuild does not preserve env variables set in Dockerfile, so need to add it to PATH here.


TerryHowe (approved):
/lgtm
/approve

PR #1047, opened 2022-07-12
Get active network interface to assign Packer HTTP server IP

abhay-krishna:
We do this to avoid hardcoding the interface name to `eth0` since on some systems, the primary network interface may be of a different name.


TerryHowe (approved):
/lgtm

PR #1069, opened 2022-07-19
schema should validate any extra configurations not available to use

a-cool-train:
*Description of changes:*
Currently the json schema doesn't validate when a user adds an unknown configuration. This fixes that issue.


ivyostosh:
Does this update apply to other packages as well?

a-cool-train:
> Does this update apply to other packages as well?

No just hello-eks-anywhere schema for now. But it will apply when other packages generate a schema.

TerryHowe (approved):
/lgtm
/approve

PR #1083, opened 2022-07-25
Add docs for updating helm charts

ivyostosh:
*Issue #, if available:* N/A

*Description of changes:* Add docs for updating helm charts

### DIFF
--- a/docs/development/update-helm-charts.md
+++ b/docs/development/update-helm-charts.md
@@ -1,0 +1,42 @@
+# Update helm charts for EKS-A Packages
+
+Helm charts used in EKS-A Packages are required to go through a series of modification to meet release standards. This file documents some standard procedures and modifications that are required for helm charts used in EKS-A Packages.
+
+## Generate patch files
+
+Helm charts modifications are done through patches. To do so, we perform the following procedures:
+- clone the target helm chart repo locally;
+- update helm charts locally (see details at next step); and 
+- generate patch files using [`git format-patch`](https://git-scm.com/docs/git-format-patch).
+
+## Update helm charts locally
+### Update `values.yaml`
+
+Following changes need to be made to the `values.yaml` file:
+
+| Field                                   | Action            | Value                                     |
+| :-------------------------------------: | :-------------:   | :----------------------------------------:|
+| `sourceRegistry`                        | add               | `public.ecr.aws/eks-anywhere`             |
+| `image:repository`                      | modify            | $(HELM_DESTINATION_REPOSITORY)[^1]        |
+| `image:tag`                             | delete            | N/A       
+| `image:digest`                          | add               | `{{`$(HELM_DESTINATION_REPOSITORY)[^1]`}}`|
+
+### Update `templates` directory
+Following changes need to be made to the yaml files under `templates`:
+
+| Field                                   | Action            |  Value                                    |
+| :-------------------------------------: | :---------------: | :---------------------------------------: |
+| `metadata:namespace`[^2]                | add               | `{{ .Release.Namespace \| quote }}`       |
+| `spec:template:spec:containers: image`  | modify            | `{{ .Values.sourceRegistry }}/{{ .Values.image.repository }}@{{ .Values.image.digest }}`        |
+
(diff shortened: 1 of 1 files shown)


docs/development/update-helm-charts.md
```diff
@@ -0,0 +1,42 @@
...
+
+Following changes need to be made to the `values.yaml` file:
+
+| Field                                   | Action            | Value                                     |
+| :-------------------------------------: | :-------------:   | :----------------------------------------:|
+| `sourceRegistry`                        | add               | `public.ecr.aws/eks-anywhere`             |
+| `image:repository`                      | modify            | $(HELM_DESTINATION_REPOSITORY)[^1]        |
+| `image:tag`                             | delete            | N/A       
+| `image:digest`                          | add               | `{{`$(HELM_DESTINATION_REPOSITORY)[^1]`}}`|
+
```
junshun, line 23: We will need to add one more row here for imagePullSecrets (I  dont believe it is in most charts by default). Name still to be finalized.

`| `imagePullSecrets` |  add  | `regcreds`|`

docs/development/update-helm-charts.md
```diff
@@ -0,0 +1,42 @@
...
+### Update `templates` directory
+Following changes need to be made to the yaml files under `templates`:
+
+| Field                                   | Action            |  Value                                    |
+| :-------------------------------------: | :---------------: | :---------------------------------------: |
+| `metadata:namespace`[^2]                | add               | `{{ .Release.Namespace \| quote }}`       |
+| `spec:template:spec:containers: image`  | modify            | `{{ .Values.sourceRegistry }}/{{ .Values.controller.image.repository }}@{{ .Values.controller.image.digest }}`        |
+
```
junshun, line 31: Along with the previous pullSecrets in values added

| `spec:template:spec:` | add | `    {{- with .Values.imagePullSecrets }} / imagePullSecrets: / {{- toYaml . | nindent 8 }} / {{- end }}` |

docs/development/update-helm-charts.md
```diff
@@ -0,0 +1,42 @@
...
+## Update helm charts locally
+### Update `values.yaml`
+
+Following changes need to be made to the `values.yaml` file:
+
+| Field                                   | Action            | Value                                     |
+| :-------------------------------------: | :-------------:   | :----------------------------------------:|
+| `sourceRegistry`                        | add               | `public.ecr.aws/eks-anywhere`             |
+| `image:repository`                      | modify            | $(HELM_DESTINATION_REPOSITORY)[^1]        |
```
TerryHowe, line 20: Maybe it would be easier to just have `### sourceRegistry` etc and break each one of these 5 items down with prose instead of table and footnotes.

docs/development/update-helm-charts.md
ivyostosh: Added

docs/development/update-helm-charts.md
ivyostosh: Added

docs/development/update-helm-charts.md
ivyostosh: Make sense. I was starting to have trouble adding more useful contexts in this table format. Updated per suggestion.

docs/development/update-helm-charts.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Update helm charts locally
+### Update `values.yaml` file
+
+Following changes need to be made to the `values.yaml` file:
+
+- `sourceRegistry`
+    - Add this field with value `public.ecr.aws/eks-anywhere`. 
+    - To test the helm chart locally, you can call `helm install` with flag `--set sourceRegistry=${YourECRRegistry}` to override its value.
+- `image:repository`
```
lewisdiamond, line 20: `image.repository` won't always be at the root. In the example you wrote, `metallb/controller` and `metallb/speaker` are in a single helm chart, not multiple. Each have their own sections where you can set the image repository. I'd instead write that the image registry and repository should be configured via `values.yaml` to allow overriding and that `digest` is set in `values.yaml` and replaced automatically by the build process.

ivyostosh: Thanks for pointing this out! I updated the example description of metallb to be more accurate. I also added (1) a note on the top of this doc that helm chart repo structure varies, and (2) modified the field descriptions to be more open-ended.

PR #1088, opened 2022-07-26
helm/build: Read manifest digest from the local docker image

ewollesen:
Previously, the manifest digest was being read from the image in its
repository. The problem with that, is that if you're building locally for the
first time, then that image hasn't been pushed to the repository yet, hence
skopeo would fail to read the manifest, or its digest. Furthermore, if the
image did exist in the repository, but it differed from the one you just built
locally, then the manifest digest was just plain incorrect.

To read the manifest from the local image, the helm Makefile rules are
modified to more closely mimic other project-specific pattern rules, like
%/images/amd64. This provides a way to pass in the project name, and from it
the path to the local tarball, from which the manifest digest is read.

I also added a more explicit check for the number of arguments passed to
helm_require.sh, because I've been bitten more than once by one of those
arguments being an empty string, which just causes the available arguments to
shift down, leaving HELM_IMAGE_LIST empty, and then the script gives strange
error messages. I might look into whether quoting the arguments to
helm_require.sh can also alleviate that problem.

*Issue #, if available:*

*Description of changes:*


Common.mk
```diff
@@ -148,8 +148,8 @@ IMAGE_IMPORT_CACHE?=type=registry,ref=$(LATEST_IMAGE) type=registry,ref=$(subst
 
 BUILD_OCI_TARS?=false
 
-LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) 
-IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push)) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) 
+LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(foreach image,$(HELM_IMAGE_LIST),$(image)/helm/build)
```
jaxesn, line 151: Doesnt need to change in this PR, but id really like to move the helm targets off the IMAGE_TARGETS list and instead probably just put them directly on the BUILD_TARGETS and RELEASE_TARGETS, or define a new HELM_TARGETS.  Its never really felt right to me for them to be included in images... if i run `make images`, am i expecting helm to also build?

jaxesn (commented):
Makefile changes seem reasonable. I cant really attest to the helm chart specific stuff, but based on what i know seems reasonable.

TerryHowe (changes requested):
There could be some minor fixes out of this we want to bring in, but generally this is not going to work because the registry image will be a multi arch oci image which will not match the single arch docker local image.

TerryHowe:
I'm closing this. The docker sha will just be for the arch specific image and we want the sha of the multi arch manifest. Devs need to build the images before the helm chart.

PR #1089, opened 2022-07-27
Fix cloud-provider-aws extra flags

TerryHowe:
I didn't detect any problem from this, but this is more correct


projects/kubernetes/cloud-provider-aws/.gitignore
```diff
@@ -0,0 +1 @@
+cloud-provider-aws
```
junshun, line 1: I believe the output of the credential provider we are building is ecr-credential-provider. Should we not be adding that to .gitignore? Or is this for ignoring the whole directory?

PR #1091, opened 2022-07-27
Use envoy with LoadBalancerClass to expose Tinkerbell services

abhinavmpandey08:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/2745, https://github.com/aws/eks-anywhere/issues/2772

*Description of changes:*
This PR does a few things:
- Exposes `hegel` and `tink-server` externally using envoy and a single load balancer service
- Uses LoadBalancerClass for the service to ensure only kube-vip handles the service
- Re-organizes the templates so they are easier to maintain and follow


abhay-krishna:
nit: why do we need the `<tink service>-` prefix for the template files if the folder name itself is `<tink service>`.

abhinavmpandey08:
> nit: why do we need the `<tink service>-` prefix for the template files if the folder name itself is `<tink service>`?

Removed the service specific prefix from the templates

projects/tinkerbell/tinkerbell-chart/chart/values.yaml
```diff
@@ -26,12 +26,13 @@ boots:
...
-  image: quay.io/tinkerbell/hegel:latest
+  image: public.ecr.aws/eks-anywhere/tinkerbell/hegel:7b286fdc8e8fa91a6e9a179a5494b6ee29fce17b-eks-a-12
   imagePullPolicy: IfNotPresent
+  replicas: 1
   port:
-    hostPortEnabled: true
+    hostPortEnabled: false
     hostPort: 50061
-  args: []
+  args: ["--grpc-use-tls=false"]
```
ptrivedi, line 35: why was this needed?

abhinavmpandey08: We override this value from the CLI so it shouldn't matter anyways
https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/stack/stack.go#L154

PR #1129, opened 2022-08-03
Metallb and Harbor imagePullSecret and defaultNamespace

junshun:
*Description of changes:*
Patch to both harbor and metallb to populate imagePullSecret with data as well as defaultNamespace to the Helm Values. The default for metallb is metallb-system to match upstream and harbor for harbor.


TerryHowe (approved):
/lgtm

PR #1147, opened 2022-08-05
Fix Raw builds on CI

vignesh-goutham:
*Description of changes:*
CI builds were failing on unbound $ARTIFACTS_BUCKET. This fixes to consume the env var as an argument instead of relying on the environment.


pokearu (approved):
/lgtm
/approve

PR #1149, opened 2022-08-08
[ImageBuilder] Extracting tarball directly to download location

vignesh-goutham:
*Description of changes:*
Instead of extracting and then moving the image-builder executable, directly extracting to location to make it neater and less error prone.


mitalipaygude (approved):

TerryHowe (approved):
/lgtm

PR #1152, opened 2022-08-09
Removing old projects

pokearu:
*Description of changes:*
These projects were removed from CI hence removing from build tooling.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1153, opened 2022-08-09
Removing old projects (#1152)

pokearu:
* Removing old projects from build tooling

* Updating projects list


abhay-krishna (commented):
/lgtm
/approve

PR #1180, opened 2022-08-17
Excludes packages from staging bundle batch build

jaxesn:
*Issue #, if available:*

*Description of changes:*

While running the staging build, the provider-aws failed because there were no codebuild/code commit setup. We are fixing the cdk side, but because of this it highlighted that we do not to be building these during the staging batch build.

*We may run into issues with this and will revert if neccessary


TerryHowe (approved):
/lgtm

PR #1215, opened 2022-08-22
Bumps cilium to 1.10.14

jaxesn:
*Issue #, if available:*

*Description of changes:*

Tested this on an existing eks-a cluster by removing the installed cilium and installing the new version manually.  I also compared the helm generated template to the upstream helm generated template and confirmed there were no unexpected diffs.

*Note* There is a 1.11.x build now and I think we will update to that version shortly, but I want to get this version bump in first to validate and then backport to release-0.11.


TerryHowe (approved):
/lgtm
/approve

PR #1222, opened 2022-08-23
Add ADOT package

ivyostosh:
*Issue #, if available:* N/A

*Description of changes:* Add otel-collector project


jaxesn (commented):
Are these images going to served in production from the ecr public uri or will they published to private ecr and be behind the gate?

ivyostosh:
> Are these images going to served in production from the ecr public uri or will they published to private ecr and be behind the gate?

@jaxesn They will eventually live in our private image ECR repo though this is not reflected in this PR. I will modify the CDK to add a ECR pull through cache rule to mirror the public image, and modify this build tooling repo one more time to point to the  private ECR. I figured I could do testing separately for those few items to keep things clean.

projects/aws-observability/aws-otel-collector/Makefile
```diff
@@ -0,0 +1,41 @@
+BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
+GIT_TAG:=$(shell cat GIT_TAG)
+HELM_GIT_TAG:=$(shell cat HELM_GIT_TAG)
+GOLANG_VERSION?="1.17"
```
jaxesn, line 4: This can probably be removed?

projects/aws-observability/aws-otel-collector/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+REPO_OWNER=aws-observability
+REPO=aws-otel-collector
+
+SIMPLE_CREATE_BINARIES=false
+SIMPLE_CREATE_TARBALLS=false
+HAS_LICENSES=false
+IMAGE_NAMES=
+
+BUILD_TARGETS=helm/build
```
jaxesn, line 17: Can we avoid these two overrides? Without this what targets are run?

projects/aws-observability/aws-otel-collector/README.md
```diff
@@ -0,0 +1,11 @@
+## **AWS Distro for OpenTelemetry Collector**
+
+AWS Distro for OpenTelemetry Collector ([ADOT Collector](https://github.com/aws-observability/aws-otel-collector)) is an AWS supported version of the upstream OpenTelemetry Collector and is distributed by Amazon. It supports the selected components from the OpenTelemetry community.
+
+### Updating
+
+1. Review [releases](https://github.com/aws-observability/aws-otel-collector/releases)
+1. Update the `GIT_TAG` file to have the new desired version based on the upstream release tags.
```
jaxesn, line 8: Should we mention the update digests?

release/staging-build.yml
```diff
@@ -83,6 +83,12 @@ batch:
       env:
         variables:
           PROJECT_PATH: projects/aws/image-builder
+    - identifier: aws_observability_aws_otel_collector
```
jaxesn, line 86: I think we can skip adding this to the staging build. There is a key you can add to the makefile so it will be ignored. Check the emissary makefile

jaxesn (changes requested):
Few thoughts on the build and makefile stuff. I don't have enough expertise to review the helm patches.

projects/aws-observability/aws-otel-collector/helm/patches/0001-image-patch.yaml
```diff
@@ -0,0 +1,128 @@
...
+diff --git a/charts/opentelemetry-collector/templates/_pod.tpl b/charts/opentelemetry-collector/templates/_pod.tpl
+index 4088f97..5960003 100644
+--- a/charts/opentelemetry-collector/templates/_pod.tpl
++++ b/charts/opentelemetry-collector/templates/_pod.tpl
+@@ -18,7 +18,7 @@ containers:
+       {{- end }}
+     securityContext:
+       {{- toYaml .Values.securityContext | nindent 6 }}
+-    image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
++    image: "{{ .Values.sourceRegistry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
lewisdiamond, line 21: This should use the image digest instead.

projects/aws-observability/aws-otel-collector/helm/patches/0002-namespace-patch.yaml
```diff
@@ -0,0 +1,193 @@
...
+   {{- with .Values.serviceAccount.annotations }}
+diff --git a/charts/opentelemetry-collector/templates/servicemonitor.yaml b/charts/opentelemetry-collector/templates/servicemonitor.yaml
+index dae68d2..005efda 100644
+--- a/charts/opentelemetry-collector/templates/servicemonitor.yaml
++++ b/charts/opentelemetry-collector/templates/servicemonitor.yaml
+@@ -3,6 +3,7 @@ apiVersion: monitoring.coreos.com/v1
+ kind: ServiceMonitor
+ metadata:
+   name: {{ include "opentelemetry-collector.fullname" . }}
++  namespace: {{ .Release.Namespace | quote }}
```
lewisdiamond, line 175: We probably want a default namespace for all those namespaced resources.

projects/aws-observability/aws-otel-collector/Makefile
ivyostosh: Yes, removed the `GOLANG_VERSION`

projects/aws-observability/aws-otel-collector/Makefile
ivyostosh: I believe they are necessary. See screenshot below without the override.

Screen Shot 2022-08-26 at 9 31 15 AM

release/staging-build.yml
ivyostosh: Done

projects/aws-observability/aws-otel-collector/helm/patches/0002-namespace-patch.yaml
ivyostosh: Added. I'm naming it `observability` for now, but we can change it later if we find a name for better fit.

projects/aws-observability/aws-otel-collector/README.md
ivyostosh: Yes, good callout. I just re-organized the `README` and added more contents. I still plan to do one more update to this file once I have all other pieces sorted out (CDK changes, update frequency, etc.)

projects/aws-observability/aws-otel-collector/Makefile
jaxesn: Ah yeah i see its because the helm targets are added to the IMAGE_TARGETS, which I want to change. this is fine for now.

projects/aws-observability/aws-otel-collector/build/update_digest.sh
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -x
+set -o errexit
+set -o nounset
+set -o pipefail
+
```
TerryHowe, line 20: this looks like a copy of build/lib/image_shasum.sh

TerryHowe (commented):
generally looks good, just wondering about shasum shell script

projects/aws-observability/aws-otel-collector/build/update_digest.sh
ivyostosh: Ah I was using `Cilium` project as a reference and didn't realize the `image_shasum.sh` did the exact same thing. I made an update to reference `build/lib/image_shasum.sh` instead.

TerryHowe (approved):
/lgtm

PR #1229, opened 2022-08-25
Add skip-checks argument to sparse-checkout for individual file support

abhay-krishna:
The new builder-base updated Git from 2.34.3 to 2.37.1, which added [this commit](https://github.com/git/git/commit/4ce504360bc3b240e570281fabe00a85027532c3) to validate that the arguments to the `set` command are directories. The attribution job with the updated builder-base failed building the bootconfig binary with the error
```bash
git -C linux sparse-checkout set tools lib/bootconfig.c include/linux/bootconfig.h --cone
fatal: 'lib/bootconfig.c' is not a directory; to treat it as a directory anyway, rerun with --skip-checks
```
This should fix it for future usecases.


abhay-krishna:
Note: This presubmit will also fail since the builder-base it uses has the older Git.

PR #1239, opened 2022-08-26
Debundle helm make targets with other image make targets for build and release

ivyostosh:
*Issue #, if available:* N/A. This is for use cases where helm charts are created without images.

*Description of changes:* Debundled helm make targets with other image make targets for `build` and `release`. This should not impact any other existing make target.

*Testing:* With this change, project `ADOT` is able to avoid build_target and release_target override. I also randomly picked project `metallb` (that has both image build and helm chart build) and made sure it could still be built with original targets.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -148,8 +148,11 @@ IMAGE_IMPORT_CACHE?=type=registry,ref=$(LATEST_IMAGE) type=registry,ref=$(subst 
 
 BUILD_OCI_TARS?=false
 
-LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(if $(filter true,$(HELM_PULL)),helm/pull,) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) 
-IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push)) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) 
+LOCAL_IMAGE_WO_HELM_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64)
+IMAGE_WO_HELM_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push))
+
+LOCAL_IMAGE_TARGETS=$(LOCAL_IMAGE_WO_HELM_TARGETS) $(if $(filter true,$(HELM_PULL)),helm/pull,) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,)
+IMAGE_TARGETS=$(IMAGE_WO_HELM_TARGETS) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,)
 
 # If running in the builder base on prow or codebuild, grab the current tag to be used when building with cgo
 CURRENT_BUILDER_BASE_TAG=$(or $(and $(wildcard /config/BUILDER_BASE_TAG_FILE),$(shell cat /config/BUILDER_BASE_TAG_FILE)),latest)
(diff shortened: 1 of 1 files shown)


Common.mk
```diff
@@ -149,8 +149,11 @@ IMAGE_IMPORT_CACHE?=type=registry,ref=$(LATEST_IMAGE) type=registry,ref=$(subst
 
 BUILD_OCI_TARS?=false
 
-LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(if $(filter true,$(HELM_PULL)),helm/pull,) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) 
-IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push)) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) 
+LOCAL_IMAGE_WO_HELM_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64)
```
jaxesn, line 152: Do we need these at all? Were you thinking of preserving the make images behavior for people who might run it locally?

ivyostosh: Yes, I'm hoping to keep `make images` the same as before as I know there are quite bit of people use this target often for testing. I'm also open to change the default behavior, but would like to hear what is the best way to proceed.

jaxesn: honestly im fine changing it. we can just tell people in the build channel and they can adjust :)

ivyostosh: Changed :)

Common.mk
```diff
@@ -570,7 +573,9 @@ endif
 #### Image Helpers
 
 ifneq ($(IMAGE_NAMES),)
-.PHONY: local-images images
+.PHONY: local-images-w-helm images-w-helm local-images images
```
jaxesn, line 576: I would probably not include these extra targets, instead just have people run helm/build or helm/push depending on what they are doing. If you think there is enough of a use case here that others will use these helper targets, its totally fine to leave them. We could always remove them later if we dont hear about anyone using them

Common.mk
```diff
@@ -717,7 +722,7 @@ help-list:
 add-generated-help-block: # Add or update generated help block to document project make file and support shell auto completion
 add-generated-help-block:
 	$(BUILD_LIB)/generate_help_body.sh $(MAKE_ROOT) "$(BINARY_TARGET_FILES)" "$(BINARY_PLATFORMS)" "${BINARY_TARGETS}" \
-		$(REPO) $(if $(PATCHES_DIR),true,false) "$(LOCAL_IMAGE_TARGETS)" "$(IMAGE_TARGETS)" "$(BUILD_TARGETS)" "$(RELEASE_TARGETS)" \
+		$(REPO) $(if $(PATCHES_DIR),true,false) "$(LOCAL_IMAGE_W_HELM_TARGETS)" "$(IMAGE_W_HELM_TARGETS)" "$(BUILD_TARGETS)" "$(RELEASE_TARGETS)" \
```
jaxesn, line 725: Does this change end up making so the generated help is the same as it was?

jaxesn (commented):
FInal questions, but otherwise this is probably good to merge.

Common.mk
ivyostosh: I talked to our package team today (as I know a few frequent users that build images and helm charts together) and they seemed to be ok with the idea to keep them as separated targets. So I have removed the additional local-images-w-helm images-w-helm targets!

Common.mk
ivyostosh: That was the goal and I confirmed that was the case by testing a few repos. 

Though by a closer look today, I believe the right thing to do is really to keep "$(LOCAL_IMAGE_TARGETS)" and "$(IMAGE_TARGETS)" here, and update all add-generated-help-block for projects since it doesn't make sense to have  helm targets in the image help block anymore. So I reverted the change I made here.

I can do a separate PR just to run through a bulk update for the add-generated-help-block if needed. Or if you think project owners can run them as they go, that work for me too.

Common.mk
```diff
@@ -148,8 +148,11 @@ IMAGE_IMPORT_CACHE?=type=registry,ref=$(LATEST_IMAGE) type=registry,ref=$(subst
 
 BUILD_OCI_TARS?=false
 
-LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64) $(if $(filter true,$(HELM_PULL)),helm/pull,) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,) 
-IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push)) $(if $(filter true,$(HAS_HELM_CHART)),helm/push,) 
+LOCAL_IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(image)/images/amd64)
+IMAGE_TARGETS=$(foreach image,$(IMAGE_NAMES),$(if $(filter true,$(BUILD_OCI_TARS)),$(call IMAGE_TARGETS_FOR_NAME,$(image)),$(image)/images/push))
+
+LOCAL_IMAGE_W_HELM_TARGETS=$(LOCAL_IMAGE_TARGETS) $(if $(filter true,$(HAS_HELM_CHART)),helm/build,)
```
jaxesn, line 154: remove these two

ivyostosh: Done

TerryHowe (approved):

PR #1242, opened 2022-08-29
Use latest tag instead of git branch

vignesh-goutham:
*Description of changes:*
Re-use existing latest tag on repo instead of git branch. This helps out with the branch name/latest tag resolving properly in all environments.

### DIFF
--- a/build/lib/eksa_releases.sh
+++ b/build/lib/eksa_releases.sh
@@ -13,21 +13,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Key for bundles manifest is environment/branch_name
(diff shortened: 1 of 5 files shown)


build/lib/eksa_releases.sh
```diff
@@ -45,9 +45,9 @@ function build::eksa_releases::get_eksa_component_asset_url() {
   local -r asset=$2
   local -r release_branch=$3
   local -r dev_release=${4-false}
-  local -r branch_name=${5-main}
+  local -r latest=${5-main}
```
jaxesn, line 48: Should these defaults have changed too?

PR #1283, opened 2022-09-09
fix cross account issue

vincentni:
*Issue #, if available:*

*Description of changes:*


projects/goharbor/harbor/patches/0004-fix-cross-account-issue-for-AWS-ECR.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/src/pkg/reg/adapter/awsecr/adapter.go b/src/pkg/reg/adapter/awsecr/adapter.go
+index 1ad6ea11c..fdc8fae01 100644
+--- a/src/pkg/reg/adapter/awsecr/adapter.go
++++ b/src/pkg/reg/adapter/awsecr/adapter.go
+@@ -49,7 +49,7 @@ func newAdapter(registry *model.Registry) (*adapter, error) {
+ 		return nil, err
+ 	}
+ 	svc, err := getAwsSvc(
+-		region, registry.Credential.AccessKey, registry.Credential.AccessSecret, registry.Insecure, nil)
++		region, registry.Credential.AccessKey, registry.Credential.AccessSecret, registry.Insecure, &registry.URL)
```
junshun, line 19: Isnt there another change needed to also use this registry.URL to pass into registry ids for the getAuthorizationToken function call

vincentni: With that change, [the following lines](https://github.com/goharbor/harbor/blob/main/src/pkg/reg/adapter/awsecr/auth.go#L116-L118) will be executed so the endpoint is explicitly set there. I tested it and it is working now.

junshun: Nice

junshun:
https://github.com/goharbor/harbor/issues/15388

Putting upstream issue here to keep track of.
We should get this change upstream if it works.

vincentni:
https://github.com/goharbor/harbor/pull/17533

The change is also upstreamed.

PR #1290, opened 2022-09-12
Separate HELM_TAG from IMAGE_TAG

ivyostosh:
HELM_TAG determines the helm chart version, and helm chart ECR repo tags. Currently, HELM_TAG is inferred from the the IMAGE_TAG. This works fine when we generate images and helm charts together in build tooling. However, in the case where we use upstream images but build helm charts on our own, we would like to make the helm tag independent from upstream so we can have full control the semantic versioning of all helm charts.

HELM_TAG will take IMAGE_TAG by default, so this change should have no impact to existing projects.


Common.mk
```diff
@@ -660,14 +661,14 @@ helm/build: $(LICENSES_TARGETS_FOR_PREREQ)
...
 helm/build: $(if $(wildcard $(MAKE_ROOT)/helm/patches),$(HELM_GIT_PATCH_TARGET),)
 	$(BUILD_LIB)/helm_copy.sh $(HELM_SOURCE_REPOSITORY) $(HELM_DESTINATION_REPOSITORY) $(HELM_DIRECTORY) $(OUTPUT_DIR)
-	$(BUILD_LIB)/helm_require.sh $(HELM_SOURCE_IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR) $(IMAGE_TAG) $(LATEST) $(HELM_IMAGE_LIST)
+	$(BUILD_LIB)/helm_require.sh $(HELM_SOURCE_IMAGE_REPO) $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR) $(IMAGE_TAG) $(HELM_TAG) $(LATEST) $(HELM_IMAGE_LIST)
 	$(BUILD_LIB)/helm_replace.sh $(HELM_DESTINATION_REPOSITORY) $(OUTPUT_DIR)
 	$(BUILD_LIB)/helm_build.sh $(OUTPUT_DIR) $(HELM_DESTINATION_REPOSITORY)
 
 # Build helm chart and push to registry defined in IMAGE_REPO.
 .PHONY: helm/push
 helm/push: helm/build
```
TerryHowe, line 670: This is probably how it should be, but what is going to be the impact on existing helm charts? If IMAGE_TAG < HELM_TAG, this might break upgrades

ivyostosh: So I have `HELM_TAG?=$(IMAGE_TAG)` in line 165. Unless a project overrides the HELM_TAG in Makefile, the HELM_TAG will remain the same as the IMAGE_TAG. This means all existing projects with helm charts will not get impacted. That being said, I think in general overriding the HELM_TAG is a bad idea unless is for a use case where you are only building helm charts (which is the use case I called out in the description).

On the point of this change may break upgrades, could you elaborate a little more? Since for projects like ADOT, it's expected for the IMAGE_TAG and HELM_TAG to be different. I would like to know if there are any concerns of this approach.

TerryHowe: Well, if we are building charts now v0.2.0 because that is the value of the IMAGE_TAG and all the sudden we switch to HELM_TAG which is say v0.1.0, upgrades would not work properly

ivyostosh: I guess it does give user options to do so, but the intention of this change is not to provide flexibilities for projects to randomly change helm tags, but rather to be independent from image tags.

In my opinion, helm tag should always be $(GIT_TAG)-$(GIT_HASH), which is what we follow now and is what we will follow after this change for all projects as well. So I don't see any chart will have a scenario where they will get a v0.1.0 tag after v0.2.0. If we really want to be strict, I can hardcode HELM_TAG to $(GIT_TAG)-$(GIT_HASH) in Common.mk, but I have concerns this will impact some use cases that I'm not aware of.

If causing user confusions is a concern, I can add some user instructions in Commom.mk to add some context. Will this help?

TerryHowe (approved):
/lgtm
/approve

PR #1294, opened 2022-09-13
Update helm_requires for special HELM_TAGs

ivyostosh:
When we have cases where the HELM_TAG is different from the IMAGE_TAG, we are pulling images from upstream. Though we pull images directly from upstream for build tooling checks (i.e. get images shasums), we will use cached images in the helm charts. Therefore we added special handling in helm_require for this scenario.

This change should have no impact on existing projects.


TerryHowe (approved):
/lgtm
/approve

PR #1301, opened 2022-09-14
Bump CAPV to v1.3.1

abhay-krishna:
Bumping CAPV to the [latest release](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/releases/tag/v1.3.1).


jaxesn:
Are there cluster spec changes we need to make on the cli side as well? We also using try and check the versions of the components capv requires and see if we need to update to match

abhay-krishna:
> Are there cluster spec changes we need to make on the cli side as well? We also using try and check the versions of the components capv requires and see if we need to update to match

There is only one change that they did, which was to separate out the `VSphereMachineTemplate` for CP and worker nodes into separate resources, and changed the approprirate refs. But we were already doing that by having a separate template for MachineDeployments so we don't need to change specs on the CLI side.

jaxesn:
sounds good, lets merge once e2e tests are stably again.

/hold

PR #1305, opened 2022-09-14
Patch cluster-api for empty taints issue

d8660091:
*Issue #, if available:*

*Description of changes:* Add the patch for cluster-api empty taints bug

## Tests
1. Manually tested the patch by 1) create [] taints cluster with cluster-api tilt, 2) create [] taints cluster with EKSA by overriding the cluster-api images using the images built by tilt 
2. `make run-target-in-docker MAKE_TARGET=binaries PROJECT=kubernetes-sigs/cluster-api` succeeded


jaxesn (approved):
awesome!

projects/kubernetes-sigs/cluster-api/patches/0001-Adding-capi-support-for-Bottlerocket.patch
```diff
@@ -1,7 +1,7 @@
-From 1a08c390a95fad1a448f41f4cb2d5187daa16ac3 Mon Sep 17 00:00:00 2001
+From 858e8c1e1389a16ef9282559fee0317c25604597 Mon Sep 17 00:00:00 2001
```
ptrivedi, line 1: Hmm, how did you apply the patches? These shouldn't change in the existing patches

d8660091: My Slack message from Abhay:
> I don't think it matters if you change that commit coz it's only going to get applied in the CI environment anyway, and it'll apply cleanly with the old commit ID too in my experience

PR #1306, opened 2022-09-14
Update ADOT helm charts

ivyostosh:
This PR (1) separated HELM_TAGs from IMAGE_TAGs for ADOT; and (2) added defaultNamespace for the helm charts.


projects/aws-observability/aws-otel-collector/Makefile
```diff
@@ -1,6 +1,9 @@
 BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
 GIT_TAG:=$(shell cat GIT_TAG)
+GIT_HASH:=$(shell git -C $(BASE_DIRECTORY) rev-parse HEAD)
```
jaxesn, line 3: We should avoid this, it's only going to be a headache later. Jonathan and I had to do something similar for autoscaler and I'm not super happy about it. Can we talk thru this some?

ivyostosh: Yes. Let me touch base with you guys tomorrow.

projects/aws-observability/aws-otel-collector/Makefile
```diff
@@ -1,6 +1,9 @@
...
 HELM_GIT_TAG:=$(shell cat HELM_GIT_TAG)
+IMAGE_TAG=$(GIT_TAG)
+HELM_TAG=$(GIT_TAG)-$(GIT_HASH)
```
TerryHowe, line 6: Area of concern here I think

projects/aws-observability/aws-otel-collector/helm/patches/0002-namespace-patch.yaml
```diff
@@ -1,193 +1,222 @@
...
-index e267336..d673a40 100644
+index e267336..2658f7b 100644
 --- a/charts/opentelemetry-collector/templates/configmap-agent.yaml
 +++ b/charts/opentelemetry-collector/templates/configmap-agent.yaml
 @@ -3,6 +3,7 @@ apiVersion: v1
  kind: ConfigMap
  metadata:
    name: {{ include "opentelemetry-collector.fullname" . }}-agent
-+  namespace: {{ .Release.Namespace | default "observability" | quote }}
++  namespace: {{ .Release.Namespace | default .Values.defaultNamespace | quote }}
```
TerryHowe, line 33: Do we have a standard across charts for this. I think in the CRD, we use `targetNamespace` and others probably use `namespace`? The name `defaultNamespace` seems a bit odd because it looks like it is actually the namespace.

ivyostosh: @junshun Steven will have more context on this. It is a standard across our existing packages today. See [metallb](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/metallb/metallb/helm/patches/0002-Adding-imagePullSecrets-and-extracting-default-names.patch) as an example.

junshun: Currently the defaultNamespace value is used here so the controller knows which namespace the chart will install to in the cases where the user does not provide a targetNamespace. This info is used so we can add the ecr-token to that namespace.

TerryHowe: Okay, defaultNamespace makes sense. We need to make sure we use that across the other projects.

projects/aws-observability/aws-otel-collector/Makefile
```diff
@@ -1,6 +1,9 @@
 BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
 GIT_TAG:=$(shell cat GIT_TAG)
+GIT_HASH:=$(shell git -C $(BASE_DIRECTORY) rev-parse HEAD)
 HELM_GIT_TAG:=$(shell cat HELM_GIT_TAG)
+IMAGE_TAG=$(GIT_TAG)
```
jaxesn, line 5: image_tag needs to match whats in git_tag since image is pulled from ecr directly

projects/aws-observability/aws-otel-collector/Makefile
jaxesn: move this to common.mk and remove GIT_HASH on line 3

projects/aws-observability/aws-otel-collector/Makefile
ivyostosh: I added a comment to explain the rationale.

projects/aws-observability/aws-otel-collector/Makefile
ivyostosh: Updated. We will avoid overwriting the GIT_HASH and HELM_TAG.

projects/aws-observability/aws-otel-collector/Makefile
ivyostosh: Done

TerryHowe (commented):
/lgtm

PR #1307, opened 2022-09-14
Changes for dev package build

jonahjon:
Signed-off-by: jonahjon <jonahjones094@gmail.com>

`GIT_TAG?=$(shell cat GIT_TAG)` allows the var to be overwritten by codebuild

`${GIT_TAG#[^0-9:main]}"` makes it so the `m` won't get chopped off if the tag is main.


vincentni (approved):
/lgtm

PR #1320, opened 2022-09-20
Cluster Autoscaler Packages schema.json

jonathanmeier5:
*Issue #, if available:*

*Description of changes:*
Adding schema to publish helm chart as part of packages bundle.

### DIFF
--- a/projects/kubernetes/autoscaler/1-23/helm/schema.json
+++ b/projects/kubernetes/autoscaler/1-23/helm/schema.json
@@ -1,0 +1,1050 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "packages.eks.amazonaws.com/autoscaler.schema.json",
+  "type": "object",
+  "default": {},
+  "title": "Root Schema",
+  "required": [
+      "affinity",
+      "autoDiscovery",
+      "autoscalingGroups",
+      "autoscalingGroupsnamePrefix",
+      "clusterAPIMode",
+      "clusterAPIKubeconfigSecret",
+      "clusterAPIWorkloadKubeconfigPath",
+      "clusterAPICloudConfigPath",
+      "clusterAPIConfigMapsNamespace",
+      "cloudConfigPath",
+      "cloudProvider",
+      "containerSecurityContext",
+      "deployment",
+      "dnsPolicy",
+      "expanderPriorities",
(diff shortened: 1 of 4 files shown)


projects/kubernetes/autoscaler/1-23/helm/schema.json
```diff
@@ -0,0 +1,151 @@
...
+
+  ],
+  "extraVolumeMounts": [
+
+  ],
+  "fullnameOverride": "",
+  "image": {
+    "sourceRegistry": "public.ecr.aws",
+    "repository": "kubernetes/autoscaler",
+    "digest": "sha256:372064f0f0ef50ee025109fff55683e50d488b8ae5fe1cb62a7bbc0fc16a64ee",
```
jonathanmeier5, line 66: I'm guessing I should pull out this specific digest, and that the packages controller will populate it with the most current version?

projects/kubernetes/autoscaler/1-23/helm/schema.json
```diff
@@ -0,0 +1,151 @@
...
+  "clusterAPIMode": "incluster-incluster",
+  "clusterAPIKubeconfigSecret": "",
+  "clusterAPIWorkloadKubeconfigPath": "/etc/kubernetes/value",
+  "clusterAPICloudConfigPath": "/etc/kubernetes/mgmt-kubeconfig",
+  "clusterAPIConfigMapsNamespace": "",
+  "cloudConfigPath": "",
+  "cloudProvider": "clusterapi",
+  "containerSecurityContext": {
+  },
+  "deployment": {
```
jonathanmeier5, line 32: For now I'm leaving all of these fields exposed while I test if that's ok. I expect that I will be able to strip down a lot of them before release.

That said, it seems to me we would rather err on the permissive than restrictive side. Does that align with packages team experience?

TerryHowe: yes

projects/kubernetes/autoscaler/1-23/helm/schema.json
TerryHowe: We'll have a specific schema.json per chart, but yeh, we shouldn't be validating a particular digest in there here

TerryHowe (approved):
/lgtm

jonathanmeier5:
Merging with @TerryHowe 's approval.

PR #1349, opened 2022-09-27
bugfix: Override existing go-licenses binaries when building

ewollesen:
This should maintain the previous behavior of deleting an existing go-licenses if its already there, but with the added bonus of not aborting if there isn't a go-licenses binary already there.

*Issue #, if available:*

*Description of changes:*


TerryHowe (approved):
/lgtm
/approve

PR #1386, opened 2022-10-05
Run release branch specific builds for kind in staging

jaxesn:
*Issue #, if available:*

*Description of changes:*

Kind is the long pole when it comes to builds in the staging batch build. This breaks the build up by release branch

*NOTE* The binaries will build and publish in each build and since they are not released branch it will overwrite, this shouldn't be an issue since the checksums will all guarantee the resulting builds are all the same, but something to keep in mind just in case.

### DIFF
--- a/release/staging-build.yml
+++ b/release/staging-build.yml
@@ -308,7 +308,7 @@ batch:
           CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/kubernetes-sigs.image-builder
           IMAGE_OS: bottlerocket
           RELEASE_BRANCH: 1-24
-    - identifier: kubernetes_sigs_kind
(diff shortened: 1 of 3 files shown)


release/staging-build.yml
```diff
@@ -308,7 +308,7 @@ batch:
           CLONE_URL: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/kubernetes-sigs.image-builder
           IMAGE_OS: bottlerocket
           RELEASE_BRANCH: 1-24
-    - identifier: kubernetes_sigs_kind
+    - identifier: kubernetes_sigs_kind_1_20_buildspec
```
abhay-krishna, line 311: Is it getting the name of the buildspec file minus the extension at the end?

PR #1389, opened 2022-10-05
Adds docker option for cgo builds

jaxesn:
*Issue #, if available:*

*Description of changes:*
Due to common build failures in codebuild we are trying to see if using docker instead of buildctl to build the cgo binaries will be more consistent.

This also removes the source-controller dep from flux since we arent building that helm chart anymore.


abhay-krishna (approved):
/lgtm
/approve

PR #1391, opened 2022-10-05
Include cert manager schema

a-cool-train:
*Description of changes:*
- Include json schema for cert manager


jonahjon:
Just a heads up, we don't have to include every field from the values.yaml in the schema. If there are things that don't make sense it's ok to drop them prior to conversation so it's not overly bloated.

a-cool-train:
> Just a heads up, we don't have to include every field from the values.yaml in the schema. If there are things that don't make sense it's ok to drop them prior to conversation so it's not overly bloated.

gotcha. That makes sense.

PR #1398, opened 2022-10-06
Specify correct appVersion in Cluster Autoscaler

jonathanmeier5:
*Issue #, if available:*

*Description of changes:*
Previously, we were not updating the chart's default appVersion of `1.23.0`.

Unfortunately, we do not have access to the specific cluster autoscaler version tags because we tie our builds to git hashes.
We have to do this because the upstream project has messed up their tagging.

This is far from an ideal solution, and I am open to alternatives.

I tried generating a Major.Minor vesion based on the RELEASE_BRANCH, but passing that variable down into helm_require.sh is non trivial.


TerryHowe (approved):
/lgtm

PR #1413, opened 2022-10-07
Add git hash to cluster autoscaler helm tags

jonathanmeier5:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/3186

*Description of changes:*
To avoid overwriting tags, we need to add the git hash of the build tooling repo at build time.


TerryHowe (approved):
/lgtm

PR #1426, opened 2022-10-13
Bump packages v.0.2.10

jonahjon:
*Issue #, if available:*

*Description of changes:*


projects/aws/eks-anywhere-packages/GIT_TAG
```diff
@@ -1 +1 @@
-v0.2.8
+v0.2.10
```
TerryHowe, line 1: should we be build v0.2.9-alpha1 or something of that sort? There is going to be at *least* one more release after this

jonahjon: I'm not sure if it matters too much since we're going to v1.0 soon, but I'm down to do that if you prefer.

jonahjon: Since we can go to v.0.3.0 next release, and then v.1.0.0 the one following that.

TerryHowe: I would disagree that we are ready for v1. Maybe after the dust settles on crd upgrades

TerryHowe (approved):
/lgtm
/approve

PR #1458, opened 2022-10-18
Fix dependencies being empty if undefined

lewisdiamond:
*Description of changes:*
Fix dependencies being generated with empty list when none are defined.

### DIFF
--- a/build/lib/helm_require.sh
+++ b/build/lib/helm_require.sh
@@ -100,7 +100,7 @@ then
 !
 fi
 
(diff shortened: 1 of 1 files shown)


jaxesn:
does this need cherrypicked?

jaxesn:
if emissary fails due to space issue, just override it

lewisdiamond:
@jaxesn I don't think this needs to be cherrypicked, it's just for packages.

lewisdiamond:
/cherry-pick release-0.12
Just to limit the diff

PR #1470, opened 2022-10-19
Fix CA & Metrics Server imagePullSecrets Schema

jonathanmeier5:
*Issue #, if available:*
Epic: https://github.com/aws/eks-anywhere/issues/3186

*Description of changes:*
The imagePullSecrets schema field was wrong before. It accepted an array of strings when it should have accepted an array of objects.

This PR resolves the issue.

There was another issue in the metrics-server helm chart where the imagePullSecrets key was under `image` instead of being a top-level key.

None of these things caused actual bugs because the controller overrides imagePullSecrets and injects it as a value to the helm chart. All the same, I'd like to make these changes to keep things tight.

I've manually verified the changes by generating a bundle locally and applying it to a cluster. With the dev bundle I've successfully applied package configs that specify `imagePullSecrets` keys with object lists.


TerryHowe (approved):
/lgtm

PR #1481, opened 2022-10-19
Standardize namespace setting and add missing on webhook

lewisdiamond:
*Description of changes:*
Standardize namespace setting across the chart (with default to what's in the values.yaml file).
Add missing namespacing to webhook service.


TerryHowe (approved):
/lgtm

TerryHowe:
all that approval and probably will fail checksums

junshun:
This patches charts not images so no checksum fails I would think

PR #1491, opened 2022-10-20
Make sourceRegistry Top Level Value

jonathanmeier5:
*Description of changes:*
Fixes a bug in cluster-autoscaler and metrics-server packages.

The previous registry key was image.sourceRegistry when it should have been sourceRegistry.

Additionally, the default image registry was set to a public AWS registry when it should have pointed to 7837 account.

Because of these two things together, the controller did not override the registry with the prod account and instead tried to pull images from the public eks-anywhere account, which did not contain the images.

Manually tested by building both packages locally and creating bundle locally. Successfully installed both packages on a dev cluster pointing to prod account with the chart image sha's overwritten to point at image shas in the production account.

Additionally verified that the new schema.json sourceRegistry field works correctly.


TerryHowe (approved):
/lgtm

PR #1528, opened 2022-11-07
Support build-your-own Ubuntu AMI through image-builder CLI

abhay-krishna:
This PR adds support for building your own AMIs via the image-builder CLI. At the moment, only Ubuntu is supported as the OS. With this change, Ubuntu AMI builds in CI will use the image-builder CLI instead of invoking Packer directly.


projects/aws/image-builder/builder/types.go
```diff
@@ -81,3 +84,17 @@ type NutanixConfig struct {
...
 }
+
+type AMIConfig struct {
+	AMIFilterName       string `json:"ami_filter_name"`
+	AMIRegions          string `json:"ami_regions"`
+	AWSRegion           string `json:"aws_region"`
+	AnsibleExtraVars    string `json:"ansible_extra_vars"`
+	BuilderInstanceType string `json:"builder_instance_type"`
+	CustomRole          string `json:"custom_role"`
+	CustomRoleNames     string `json:"custom_role_names"`
```
g-gaston, line 95: Maybe we could make this an array? Imagebuilder will take it as space separated collection (I think), but there is not reason we need to expose that to the user, right?
Having an array here will make more clear they can specify more than one custom role

abhay-krishna: Yes you're right, image-builder does take it as a space or comma separated string. The reason I did this is because the user input file is directly used as the packer var-file for image-builder. All we do is unmarshal the input config file, validate the fields and marshal it to the canonical packer var file location  I can see if some transformation in the pipeline is a possible option.

abhay-krishna: To maintain type parity with upstream Imagebuilder, I added another parameter custom_role_names_list which is an array. Users can use both the array field or the space-separated field, but there is a check to ensure they're mutually exclusive. If user provides the list, it is converted to the delimited string format and assigned to the the above field for Marshaling into a format that Packer can use directly with upstream image-builder. I can make it clear in the docs how these fields are supposed to be used.

projects/aws/image-builder/cmd/build.go
```diff
@@ -155,12 +156,47 @@ func ValidateInputs(bo *builder.BuildOptions) error {
...
+				return err
+			}
+			if bo.AMIConfig.CustomRole == "true" {
+				if (bo.AMIConfig.CustomRoleNameList == nil && bo.AMIConfig.CustomRoleNames == "") ||
+					(bo.AMIConfig.CustomRoleNameList != nil && bo.AMIConfig.CustomRoleNames != "") {
+					log.Fatalf("Exactly one of \"custom_role_name_list\" or \"custom_role_names\" must be provided")
+				}
+
+				if bo.AMIConfig.CustomRoleNameList != nil {
+					bo.AMIConfig.CustomRoleNames = strings.Join(bo.AMIConfig.CustomRoleNameList, " ")
```
g-gaston, line 170: you probably want to nullify `CustomRoleNameList` here so it doesn't get marshaled back?

projects/aws/image-builder/cmd/build_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"testing"
+
+	"github.com/aws/eks-anywhere-build-tooling/image-builder/builder"
+)
+
+func TestValidateSupportedHypervisor(t *testing.T) {
+	testCases := []struct {
+		testName     string
+		buildOptions builder.BuildOptions
+		wantErr      bool
```
g-gaston, line 13: I always find more useful to include the error message in the test case
That way to verify you are returning an error where you expect it and not somewhere else

g-gaston (commented):
Two minor things, otherwise looks good to me
If you are in a rush, you can address them in a separate PR, let me know

g-gaston (approved):

PR #1541, opened 2022-11-09
Update make add-generated-help-block and attribution-files logic

ivyostosh:
While evaluating the PROJECT_PATH, instead of replacing all `_` with `/`, we will only replace the first instance of `_`. This is to help support project that contains `_` in the project name.


abhay-krishna:
Thanks @ivyostosh!
/lgtm
/approve

PR #1546, opened 2022-11-10
Adds support to running just arm binary validation for source-controller

jaxesn:
*Issue #, if available:*

*Description of changes:*

Now that we have arm nodes on the postsubmit cluster, we can build the source controller arm binary in presubmits. This adds the ability to ignore the amd validation like we were already doing for arm.  There will be two presubmits for source controller, one for amd and one for arm


abhay-krishna:
Sick
/lgtm
/approve

PR #1557, opened 2022-11-11
Update hello-eks-anywhere json schema

ivyostosh:
In order for the json schema to be usable on the helm level, we need to add in additional fields.

### DIFF
--- a/projects/aws-containers/hello-eks-anywhere/helm/schema.json
+++ b/projects/aws-containers/hello-eks-anywhere/helm/schema.json
@@ -4,10 +4,32 @@
   "title": "hello-eks-anywhere",
   "type": "object",
   "properties": {
+    "namespace": {
+      "type": "string",
+      "default": "default",
+      "description": "Namespace for package and resources."
+    },
     "sourceRegistry": {
       "type": "string",
       "default": "public.ecr.aws/eks-anywhere",
       "description": "Source registry for package."
+    },
+    "repository": {
+      "type": "string",
+      "default": "hello-eks-anywhere",
+      "description": "Repository for package."
+    },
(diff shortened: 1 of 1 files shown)


projects/aws-containers/hello-eks-anywhere/helm/schema.json
```diff
@@ -4,11 +4,33 @@
...
+      "default": [],
+      "description": "Container imagePullSecrets.",
+      "items": {
+        "type": "object"
+      }
+    },
+    "tag": {
+      "type": "string",
+      "description": "Container image tag."
+    },
```
TerryHowe, line 33: Not sure we'd want to expose this or repository. Kind of goes against what were are trying to do here. No doubt we do in other packages, but I would argue we shouldn't. With either of these in there, someone that only had permissions to create packages resources could easily get access to other resources.

projects/aws-containers/hello-eks-anywhere/helm/schema.json
```diff
@@ -4,11 +4,33 @@
   "title": "hello-eks-anywhere",
   "type": "object",
   "properties": {
+    "namespace": {
+      "type": "string",
+      "default": "default",
+      "description": "Namespace for package and resources."
+    },
```
TerryHowe, line 11: I think adding namespace here would be confusing because we aren't going to use it from the package `config` we'll use `targetNamespace`

TerryHowe (changes requested):
I think we should set `"additionalProperties": true`

ivyostosh:
We will merge https://github.com/aws/eks-anywhere-build-tooling/pull/1563 instead

PR #1566, opened 2022-11-14
Switch release branched builds to batch for codebuild

jaxesn:
*Issue #, if available:*

*Description of changes:*

We already switched these builds to separate build instances for the staging build.  When bumping eks-d versions which triggers a number of builds, these builds end up causing a bunch of noisy, false-negatives because they take too long to finish.  This should allow us to break these up.

Note a CDK change will be required.


abhay-krishna (approved):
/approve

PR #1567, opened 2022-11-14
Add Prometheus helm chart

ivyostosh:

### DIFF
--- a/projects/prometheus/prometheus/helm/patches/0002-image-patch.diff
+++ b/projects/prometheus/prometheus/helm/patches/0002-image-patch.diff
@@ -1,0 +1,63 @@
+From 9a00035a01b22efd07809f5a2392c3742e985ac4 Mon Sep 17 00:00:00 2001
+From: "Ostosh, Ivy" <ivyjin215@gmail.com>
+Date: Mon, 14 Nov 2022 10:09:11 -0600
+Subject: [PATCH] Update image repo
+
+---
+ charts/prometheus-server/templates/server/deploy.yaml | 2 +-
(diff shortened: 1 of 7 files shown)


projects/prometheus/prometheus/helm/patches/0002-image-patch.diff
```diff
@@ -0,0 +1,63 @@
...
+diff --git a/charts/prometheus-server/templates/server/deploy.yaml b/charts/prometheus-server/templates/server/deploy.yaml
+index 3f327376..bbf8ce29 100644
+--- a/charts/prometheus-server/templates/server/deploy.yaml
++++ b/charts/prometheus-server/templates/server/deploy.yaml
+@@ -53,7 +53,7 @@ spec:
+       {{- end }}
+       containers:
+         - name: {{ template "prometheus.name" . }}-{{ .Values.server.name }}
+-          image: "{{ .Values.server.image.repository }}:{{ .Values.server.image.tag }}"
++          image: "{{ .Values.sourceRegistry }}/{{ .Values.server.image.repository }}@{{ .Values.server.image.digest }}"
```
TerryHowe, line 21: This is fine, but I've been thinking we should perfect the helper template and attempt to upstream so it supports tag or digest

TerryHowe (approved):
/lgtm

projects/prometheus/prometheus/helm/patches/0002-image-patch.diff
ivyostosh: I created https://github.com/aws/eks-anywhere-packages/issues/616 to track this for now. Ideally, we should be doing this for all packages.

PR #1578, opened 2022-11-16
Set default input values for AMI build configuration

abhay-krishna:
Setting default opinionated values for AMI build configuration so customers can set only values for parameters they really want to control while letting the tool default to these values for other fields.

### DIFF
--- a/projects/aws/image-builder/cmd/build.go
+++ b/projects/aws/image-builder/cmd/build.go
@@ -157,7 +157,22 @@ func ValidateInputs(bo *builder.BuildOptions) error {
 				}
 			}
 		case builder.AMI:
-			if err = json.Unmarshal(config, &bo.AMIConfig); err != nil {
+			// Default configuration for AMI builds
+			defaultAMIConfig := &builder.AMIConfig{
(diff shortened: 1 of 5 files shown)


projects/aws/image-builder/cmd/build.go
```diff
@@ -8,9 +8,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/spf13/cobra"
-
 	"github.com/aws/eks-anywhere-build-tooling/image-builder/builder"
+	"github.com/spf13/cobra"
```
jiayiwang7, line 12: nit: pkg import order

projects/aws/image-builder/cmd/build_test.go
```diff
@@ -3,16 +3,15 @@ package cmd
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
 	"github.com/aws/eks-anywhere-build-tooling/image-builder/builder"
+	"github.com/stretchr/testify/assert"
```
jiayiwang7, line 7: nit: order

jiayiwang7 (approved):

PR #1601, opened 2022-11-22
Update BR version to v1.11.0

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Update BR version to v1.11.0

### DIFF
--- a/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES
+++ b/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES
@@ -1,14 +1,14 @@
 1-20:
     ova-release-version: v1.10.1
 1-21:
-    ova-release-version: v1.10.1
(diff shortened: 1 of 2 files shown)


projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES
```diff
@@ -1,14 +1,14 @@
 1-20:
-    ova-release-version: v1.10.1
+    ova-release-version: v0.10.1
```
jaxesn, line 2: This looks wrong?

mitalipaygude: Yeah i was keeping the version for 1.20 because BR deprecated 1.20 looks like but then we would need different versions of the admin container image for the different BR versions since we wont be deprecating 1.20 till the December minor release.

PR #1618, opened 2022-11-30
Deprecate 1.20

mitalipaygude:
*Issue #, if available:*
Deprecate 1.20


TerryHowe (approved):
/lgtm

Seems to be more deleting than deprecating

PR #1630, opened 2022-12-06
Add build info to eksa-packages binaries.

lewisdiamond:
Requires latest pkg controller code.

/hold


TerryHowe (approved):
/lgtm

projects/aws/eks-anywhere-packages/Makefile
```diff
@@ -1,6 +1,12 @@
 BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
+PROJECT_COMMIT_HASH:=$(shell git -C eks-anywhere-packages/ rev-parse --short HEAD)
+TIME:=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
```
jaxesn, line 3: We tend to not add times to our builds since it will break reproducibility and the checksum validation. Usually we would use the time from the commit.

lewisdiamond: Yea I can use the ts of the commit. That makes sense.

jaxesn (approved):

PR #1673, opened 2022-12-21
Patch locale disparities by explicitly erroring for unexpected locales

chrisdoherty4:
Locales impact file ordering among other things for programs that are locale aware (most GNU programs are). Consequently, if 2 environments use different locales its possible generated files are inconsistent. This can cause the CI to fail unexpectedly without the ability to reproduce locally.

This change ensures we error with a clear message indicating how to create determinisim.


TerryHowe (approved):
/lgtm

PR #1691, opened 2022-12-29
Patch locale warning so it only shows when relevant

chrisdoherty4:
Patch the locale warning so it only shows when its known to cause an issue, namely on Linux.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -404,10 +404,13 @@ RELEASE_TARGETS?=validate-checksums $(if $(IMAGE_NAMES),images,) $(if $(filter t
 # generate files that are subsequently validated by the CI. If local environments use different 
 # locales to the CI we get unexpected failures that are tricky to debug without knowledge of 
 # locales so we'll explicitly warn here.
-ifneq ($(call TO_LOWER, $(LANG)), c.utf-8)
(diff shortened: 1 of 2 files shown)


Common.mk
```diff
@@ -404,10 +404,21 @@ RELEASE_TARGETS?=validate-checksums $(if $(IMAGE_NAMES),images,) $(if $(filter t
...
 # locales to the CI we get unexpected failures that are tricky to debug without knowledge of 
 # locales so we'll explicitly warn here.
-ifneq ($(call TO_LOWER, $(LANG)), c.utf-8)
-ifneq ($(call TO_LOWER, $(LANG)), posix)
-  $(warning WARNING: Environment locale set to $(LANG). This may create non-deterministic behavior when generating files. If the CI fails validation try `LANG=C.UTF-8 make <recipe>` to generate files instead.)
-endif
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
+  LOCALE_VALID := 0
+  ifeq ($(call TO_LOWER, $(LANG)),c.utf-8)
```
jaxesn, line 410: you could do something like this if you wanted:

```
ifneq ($(find $(call TO_LOWER, $(LANG))),c.utf-8 posix),)
```

PR #1719, opened 2023-01-09
Bumps CAPC to new commit with bug fix

wanyufe:
*Issue #, if available:*
CAPC has bug fix of adding zoneID when retrieving serviceOffering and diskOffering.
*Description of changes:*
Refer to new CAPC repo commmit hash


UPSTREAM_PROJECTS.yaml
```diff
@@ -130,7 +130,7 @@ projects:
           - tag: v1.2.0
       - name: cluster-api-provider-cloudstack
         versions:
-          - tag: v0.4.9-rc3
+          - commit: cefafbebf46668a36a690e076d25a35f6d24ff3f
```
jiayiwang7, line 133: can we use a git tag instead of commit?

pmotyka: for this fix we may have to go with commit hash unfortunately

lewisdiamond:
This breaks the cloudstack provider, the commit hash cannot be used directly. Can a tag be created instead?

PR #1726, opened 2023-01-10
Update cluster-api to v1.3.3

g-gaston:
## Description of changes

* Update cert-manager to v1.11.0 since the new capi uses v1.11.0. This required to add a patch to modify cert-manager's go.mod. Once of its dependencies suffers from a checksum mismatch error that prevents the controller from being built. We already contributed the required change upstream (in fact the patch comes directly from the upstream commit), but they haven't cut a new tag yet. Since we prefer to build from tags instead of commits, I just added it as a patch. We can remove this once cert-manager cuts a new tag.
* CAPI now uses go 1.19
* Dropped capi patch for kubeadm taints since that's now merged upstream.
* CAPI dropped the LICENSE file for the test module (the one that includes capd) in favor of the top level one, so we now reuse the main one for both modules.


projects/cert-manager/cert-manager/Makefile
```diff
@@ -47,7 +47,7 @@ create-manifests:
 
 .PHONY: update-cert-manager-manifest
 update-cert-manager-manifest:
-	wget -q https://github.com/cert-manager/cert-manager/releases/download/$(GIT_TAG)/cert-manager.yaml -O build/cert-manager.yaml
+	wget -q https://github.com/cert-manager/cert-manager/releases/download/$(GIT_TAG)/cert-manager.yaml -O manifests/cert-manager.yaml
```
a-cool-train, line 50: Is this necessary? If you make this change, you might want to update the README to reflect this change.

a-cool-train:
Reviewed the cert-manager upgrade and the patches look good. I added one nit comment.

projects/cert-manager/cert-manager/Makefile
g-gaston: Updated readme :)

jaxesn (approved):

PR #1741, opened 2023-01-13
Bump vsphere csi to 2.7.0

lewisdiamond:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/4626

*Description of changes:*

/hold


TerryHowe:
Closes: https://github.com/aws/eks-anywhere/issues/4626

jaxesn:
I think we tried this before and there is more to it due to new CRDs maybe? Weve been kind of holding off updating this for now while we figure out if we want to keep it or remove it altogether and let customers self manage it. We already have the ability for customers to disable it and self manage it for cases like this where they need to do more than we currently allow configuration wise and/or use newer versions.

lewisdiamond:
More work is needed to be able to upgrade the CSI driver.

PR #1755, opened 2023-01-18
Breaks part source controller into batch build

jaxesn:
*Issue #, if available:*

*Description of changes:*

We have had a lingering problem when building source-controller in codebuild due to the fact that it requires CGO and therefore an ARM based container/instance.  Up until now we have used docker (first started with buildctl) to do the `go build` for the ARM binary in an ARM based builder-base container.  This seems to work all the time in standard Codebuild builds, and in the prow checksum periodic using buildctl, but when using a Codebuild batch build we get failures about every other time.  The failure has been tough to diagnosis since the `go build` call just seems to stall.

Last week I added a `timeout retry` around our docker run call which does the arm build and this at least helped stabilize the staging build this week by "detecting" the stall out and rerunning.

@abhay-krishna and I were looking at our codebuild options and realized we could change the instance arch per tasks in the batch build.  This PR changes the source-controller build to be more line the kind build where it itself is a batch build where the amd and arm binary builds are run as separate tasks, on separate instance types, then the images build `depend_on` those tasks before kicking off.  The kind build builds all the binaries even in the image builds, its mostly separated out by release branch, whereas for source-controller code has been added to download the binaries from the s3 bucket before building images.  This required a bit of funkyness, but for the most part it was just reuse of targets and code we already had in the common.mk.

Since we are uploading the artifact tarballs from separate builds, I had to tweak the s3 sync command to not delete files in s3 that arent local, to avoid removing the other builds uploads.


build/lib/common.sh
```diff
@@ -86,7 +87,12 @@ function build::common::upload_artifacts() {
     # 1. To proper path on s3 with buildId-githash
     # 2. Latest path to indicate the latest build, with --delete option to delete stale files in the dest path
     aws s3 sync "$artifactspath" "$artifactsbucket"/"$projectpath"/"$buildidentifier"-"$githash"/artifacts --acl public-read
-    aws s3 sync "$artifactspath" "$artifactsbucket"/"$projectpath"/"$latesttag" --delete --acl public-read
+
+    if [ "$do_not_delete" = "true" ]; then
+      aws s3 sync "$artifactspath" "$artifactsbucket"/"$projectpath"/"$latesttag" --acl public-read
+    else
+      aws s3 sync "$artifactspath" "$artifactsbucket"/"$projectpath"/"$latesttag" -delete --acl public-read
```
abhay-krishna, line 94:
```suggestion
      aws s3 sync "$artifactspath" "$artifactsbucket"/"$projectpath"/"$latesttag" --delete --acl public-read
```

build/lib/generate_staging_buildspec.sh
```diff
@@ -69,27 +65,43 @@ for project in "${PROJECTS[@]}"; do
...
         for var in "BUILDSPEC_DEPENDS_ON_OVERRIDE" "BUILDSPEC_$((( $i + 1 )))_DEPENDS_ON_OVERRIDE"; do
             BUILDSPEC_DEPENDS_ON="$(make --no-print-directory -C $PROJECT_PATH var-value-$var RELEASE_BRANCH=$RELEASE_BRANCH 2>/dev/null)"
+            HARDCODED_DEP="false"
             if [[ "none" = "$BUILDSPEC_DEPENDS_ON" ]]; then
                 PROJECT_DEPENDENCIES=""
             elif [[ -n "$BUILDSPEC_DEPENDS_ON" ]]; then
+                HARDCODED_DEP="true"
                 PROJECT_DEPENDENCIES=$BUILDSPEC_DEPENDS_ON
-            fi
+            fi
```
abhay-krishna, line 83: nit: trailing whitespace?

PR #1758, opened 2023-01-20
Sets compute type for arm code build instances

jaxesn:
*Issue #, if available:*

*Description of changes:*

In the staging build we default to the giant instance type for amd which does not exist for arm. We have to specifically set the arm compute type to the normal large.


abhay-krishna (approved):
/approve

PR #1761, opened 2023-01-20
Add build logic for Bottlerocket AWS variants

abhay-krishna:
This PR adds the Bottlerocket AWS variants to build/release CI.

### DIFF
--- a/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES
+++ b/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES
@@ -1,12 +1,16 @@
 1-21:
+    ami-release-version: v1.10.1
     ova-release-version: v1.10.1
     raw-release-version: v1.10.1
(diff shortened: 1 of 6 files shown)


build/lib/validate_artifacts.sh
```diff
@@ -35,6 +35,11 @@ if [ -n "$IMAGE_FORMAT" ]; then
   fi
 fi
 
+if ! stat $EXPECTED_FILES_PATH &> /dev/null; then
```
jaxesn, line 38: this is risky, if we forget to add the file it will go unnoticed....

abhay-krishna: Right. I'll try to skip it just for this case

abhay-krishna: Yeah I fixed it, it was a silly thing

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #1806, opened 2023-02-02
Patch image-builder to add IMDS configuration support for Packer builds

abhay-krishna:
IMDS configuration options are available for the Packer Amazon EBS builder instance but not exposed as part of upstream image-builder's Packer config. So we need to patch it to make it configurable.


jaxesn (approved):

jonathanmeier5 (approved):
/lgtm
/approve

PR #1809, opened 2023-02-03
Pin kustomize version to v4.5.7

abhay-krishna:
[Kustomize](https://github.com/kubernetes-sigs/kustomize) released a new version v5.0.0 today which caused some issues in generating manifests, specifically setting `creationTimestamp` to `"null"` (with quotes). So pinning to an older version to avoid this. 


abhinavmpandey08 (approved):
/lgtm
/approve
/woof

PR #1823, opened 2023-02-08
Add 1-26 as a supported release branch

taneyland:
*Issue #, if available:*

*Description of changes:*
adds 1-26 as a supported branch.
adds 1-26 metrics server

added the target to clean-output to free up space when running presubmits for the diff versions


projects/kubernetes-sigs/image-builder/Makefile
```diff
@@ -48,6 +48,7 @@ ADDITIONAL_PAUSE_1-22_FROM=1-21
 ADDITIONAL_PAUSE_1-23_FROM=1-22
 ADDITIONAL_PAUSE_1-24_FROM=1-23
 ADDITIONAL_PAUSE_1-25_FROM=1-24
+ADDITIONAL_PAUSE_1-26_FROM=1-26
```
taneyland, line 51: Pause image tags match in links below for 1.26, so this does not need to reference the previous tag.

1. https://github.com/kubernetes/kubernetes/blob/v1.26.0/cmd/kubeadm/app/constants/constants.go#L420
2. https://github.com/kubernetes/kubernetes/blob/v1.26.0/build/pause/Makefile#L20

So I wasn't going to add this line, but the image builder presubmit fails without it

jaxesn: that makes sense. Id just leave it

projects/aws/bottlerocket-bootstrap/buildspecs/batch-build.yml
```diff
@@ -42,3 +42,8 @@ batch:
       env:
         variables:
           RELEASE_BRANCH: 1-25
+    - identifier: aws_bottlerocket_bootstrap_1_26
```
jaxesn, line 45: I think im a bit surprised by this. If we are skipping it, do we still want it to be included in the build specs?  I probably wrote the code that handles this, but dont remember if it is going to matter...

Also, we are skipping this one?

EKSD_LATEST_RELEASES
```diff
@@ -23,4 +23,8 @@ releases:
 - branch: 1-25
   number: 5
   kubeVersion: v1.25.6
+- branch: 1-26
+  number: 1
+  kubeVersion: v1.26.0
+  dev: true
 latest: 1-22
```
jaxesn, line 30: We need to have this conversation again about bumping this up...

projects/kubernetes/cloud-provider-vsphere/Makefile
```diff
@@ -6,6 +6,7 @@ REPO=cloud-provider-vsphere
 REPO_OWNER=kubernetes
 
 BASE_IMAGE_NAME?=eks-distro-minimal-base
+SKIPPED_K8S_VERSIONS=1-26
```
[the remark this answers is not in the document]
taneyland: yeah there isnt anything from capv on 1.26 yet

EKSD_LATEST_RELEASES
taneyland: i figured we could add the dev version now & we should have the 1.26 prod version before the next release so it wont be pulling from dev in the bundle

projects/aws/bottlerocket-bootstrap/buildspecs/batch-build.yml
taneyland: yeah i skipped this one too since we wont have BR 1.26 builds for a while anyway & the presubmit was showing some odd errors: https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-anywhere-build-tooling/1823/bottlerocket-bootstrap-tooling-presubmit/1623693594554535936

but this was before i added the 1..26 additional pause line for the image builder. not sure if that would affect this though

UPSTREAM_PROJECTS.yaml
```diff
@@ -123,6 +123,7 @@ projects:
           - tag: v1.23.1
           - tag: v1.24.1
           - tag: v1.25.0
+          - tag: v1.25.0
```
taneyland, line 126: Duplicated during generation bc we are using this tag for 1.25 and 1.26

projects/kubernetes/cloud-provider-vsphere/1-26/GIT_TAG
```diff
@@ -0,0 +1 @@
+v1.25.0
```
jaxesn, line 1: This is going to mean that the 1.25 build and the 1.26 are going to push to the same image tag. Its going to be the same content so I cant think of a reason it will matter but just calling it out. cc @abhay-krishna

abhay-krishna (approved):
/approve

PR #1830, opened 2023-02-09
Patch boots deployment strategy to 'Recreate'

chrisdoherty4:
Boots binds to host interfaces. If 2 instances try to bind to the same address/interface/port combination it errors so we need a recreate strategy to ensure the pod is torn down before the new one comes up.


jacobweinstock (approved):

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #1832, opened 2023-02-10
Add 1-26 autoscaler

taneyland:
*Issue #, if available:*

*Description of changes:*


projects/kubernetes/autoscaler/1-26/CHECKSUMS
```diff
@@ -0,0 +1,2 @@
+77a6a3036ba54a950203613be646f7f014bfc730fea16801ccca03021741520b  _output/1-25/bin/autoscaler/linux-amd64/cluster-autoscaler
```
jonathanmeier5, line 1: These need to point to `_output/1-26`.

jonathanmeier5 (commented):
One small required change I believe.

Additionally, are we able to run add and run 1-26 e2e tests yet? If we don't test this locally, I'd like to make sure there is an e2e test we can run against.

TerryHowe (approved):
/lgtm

PR #1847, opened 2023-02-14
Cleanup repo only in non-Codebuild environments

abhay-krishna:
If the build is running in Codebuild, there is no cloned repo so we don't need to run the cleanup.


TerryHowe (approved):
/lgtm

PR #1853, opened 2023-02-14
Update harbor-scanner-trivy to v0.30.7

vincentni:


ivyostosh (commented):
Make sure to update versions in https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/aquasecurity/harbor-scanner-trivy/README.md as well

PR #1887, opened 2023-02-27
Add more validation to ebs-init

nkvetsinski:
EBS init to skip deleted/renamed files and directories.

*Issue #, if available:* none

*Description of changes:* We noticed `nil pointer dereference` in cases when ebs-init tries to operate on file that was deleted or renamed. In the initial commit we didn't check the error from `DirEntry.Info()` which indicates if the file is present. 

Also adding nil check for `direEntry`.


vignesh-goutham (approved):
/ok-to-test
/approve

vignesh-goutham (approved):
/approve

PR #1904, opened 2023-03-01
Use codebuild batch build for checksum updates

jaxesn:
*Issue #, if available:*

*Description of changes:*

Currently we run a periodic job every night to regenerate the checksums.  This was originally added to ensure there are no changes we arent expecting.  We have also recently started using this job, via a manual trigger, to generate new checksums when changing golang versions.  This job was taking over 1.5hrs in prow and having space issues due to the lack of a overlayfs and the source-controller arm build needing to use buildctl.

This PR adds a new checksums batch buildspec, which is generated with the same script we use to generate the other batch codebuild buildspecs.  This batch build includes all (technically most) projects in this repo and just calls the `checksums` target.

The main change here is to the create_pr script which traditionally has always been used at the end of the job and only once.  Based on that assumption it was always just force pushing to the bot's branch.  To run these builds in parallel it needed to be able to handle concurrent git commits/pushes. It was changed to handle this case by optimistically pushing to the upstream branch and on failures fetching and rebasing, then trying the push again.  The same basic approach was taken for creating the PR via the github cli.  This was slightly different and includes some random waits just tot ry and avoid hitting the github api too much, since we can get rate limited there.

I tested this using a manual run of the codebuild staging bundle build. Here is an ex [PR](https://github.com/aws/eks-anywhere-build-tooling/pull/1920) it opened.  Note, I intentionally used an older builder-base to make sure there were actual checksum changes.  This new batch build takes less 15 mins, down from 1.5hrs.

Going forward the prow periodic will (has been) turned off and there will not be a nightly anymore for this process. These checksums should not be changing outside of golang changes so I think the nightly is not needed. Instead a new codebuild job will be created and can/will be manually triggered during golang/builder-base updates to get the new checksums.

TODO

- [ ] CDK changes - @abhay-krishna 


build/lib/setup.sh
```diff
@@ -19,7 +19,7 @@ if [ ! -d "/root/.docker" ]; then
     mkdir -p /root/.docker
 fi
 
-mv config/docker-ecr-config.json /root/.docker/config.json
+cp config/docker-ecr-config.json /root/.docker/config.json
```
jaxesn, line 22: the mv was causing this to show as a git diff which complicated the rebase, the cp shouldnt make a difference.

build/update-attribution-files/create_pr.sh
```diff
@@ -52,34 +52,109 @@ if [ "$(git stash list)" != "" ]; then
     build::common::echo_and_run git status
 fi
 
+function commit_push()
```
jaxesn, line 55: this is really the meet of the changes, this function is wrapped with a retry.

checksums-buildspec.yml
```diff
@@ -0,0 +1,17 @@
...
+    GITHUB_TOKEN: "github-eks-distro-pr-bot:github-token"
+
+phases:
+  pre_build:
+    commands:
+    - ./build/lib/setup.sh
+
+  build:
+    commands:
+    - echo "CHECKSUMS -text merge=union" > .gitattributes
```
jaxesn, line 14: this is a git trick to force the rebase to always succeed for checksum files by just doing the union of the diff. This only comes into play for the source controller builds which build amd and arm separately and therefore end up changing the same file.

build/lib/common.sh
```diff
@@ -384,7 +383,6 @@ function retry_with_timeout() {
         # multiple the numeric part of the timeout by 1.5 and suffix with the last char which is the unit
         TIMEOUT=$((${TIMEOUT:0:-1} * 3/2))${TIMEOUT: -1}
         echo "Command failed. Attempt $n/$max with timeout ${TIMEOUT}:"
-        delay=$(( delay * 2 ))
```
jaxesn, line 387: This was too aggressive for the codebuild checksum case where we expect lots of retries.  May bring this back in the future if we need to.

abhay-krishna (approved):

PR #1921, opened 2023-03-02
upgrade harbor to v2.7.1

vincentni:


TerryHowe (approved):
/lgtm

ivyostosh (commented):
[README](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/goharbor/harbor/README.md) needs to be updated

projects/goharbor/harbor/Makefile
```diff
@@ -1,10 +1,10 @@
 BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
 GIT_TAG:=$(shell cat GIT_TAG)
-GOLANG_VERSION?="1.17"
+GOLANG_VERSION?="1.19"
 REGISTRYVERSION=2.8.1-patch-redis
 NOTARYVERSION=v0.6.1
-TRIVYVERSION=v0.28.1
-TRIVYADAPTERVERSION=v0.29.0
+TRIVYVERSION=v0.37.3
+TRIVYADAPTERVERSION=v0.30.7
```
ivyostosh, line 7: Is dependency set up so that these two projects will be built first? Or we need to watch out for the build sequence as well?

vincentni: The build sequence is handled by our build-tooling, and versions here are simply for harbor to show them in the harbor web portal.

PR #1949, opened 2023-03-08
Small changes to kind to support iptables-nft

jaxesn:
*Issue #, if available:*

*Description of changes:*

We recently added the iptables-wrapper to the iptables minimal image so we can support iptables-nft as well as iptables-legacy. This changes the kindnetd image to use this new wrapper alternative.

I made a couple small changes to the kind build to facilitate to my testing of this and for next time we need to run similar tests.

I had thought I would need to make changes to the kind image as well to use the wrapper, however turns out we can't use the wrapper for that case. The kind [entrypoint](https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/usr/local/bin/entrypoint#L364) actually uses a version of the old way of detecting iptables mode by counting the number of iptables rules.  I do not think they can use the new style of looking for the hint created by the kubelet, since the kind vm is booting before kubelet has run. Because of this, I have left this entrypoint alone which has actually always properly switched the mode.

I have tested on both a ubuntu 22.04 node which uses nft and a al2 using legacy to ensure that all three of the following properly switch to nft vs legacy:

- kind vm node
- kube-proxy pods
- kindnetd pod

### DIFF
--- a/projects/kubernetes-sigs/kind/Makefile
+++ b/projects/kubernetes-sigs/kind/Makefile
@@ -73,6 +73,7 @@ $(GATHER_LICENSES_TARGETS): | $(FIX_LICENSES_KINDNETD_TARGET)
 s3-artifacts: $(ARTIFACTS_PATH)/manifests/kindnetd/$(GIT_TAG)/kindnetd.yaml
 
 images: kind-base-versioned/images/push kind-node/images/push
+create-ecr-repos: kind-node/create-ecr-repo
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -17,9 +17,12 @@ PROJECT_PATH_MAP=projects/$(patsubst $(firstword $(subst _, ,$(1)))_%,$(firstwor
 # generate files that are subsequently validated by the CI. If local environments use different 
 # locales to the CI we get unexpected failures that are tricky to debug without knowledge of 
 # locales so we'll explicitly warn here.
-TO_LOWER = $(shell echo $(1) | tr '[:upper:]' '[:lower:]')
+# In a AL2 container image (like builder base), LANG will be empty which is equilvant to posix
```
jaxesn, line 20: syncing this what we put in common.mk to avoid the confusing log output

g-gaston (approved):

PR #1974, opened 2023-03-10
update sedfile for harbor helm chart

vincentni:


projects/goharbor/harbor/helm/sedfile.template
```diff
@@ -1,4 +1,4 @@
-s/^version: 1.9.1$/version: ${IMAGE_TAG}/
+s/^version: 1.11.1$/version: ${IMAGE_TAG}/
```
ivyostosh, line 1: Will this work? So you don't need to hardcode it every time
`s/version: .*$/version: ${IMAGE_TAG}/`

vincentni: It gets passed by our script

PR #1989, opened 2023-03-14
Skip show-join-command in kubeadm call

taneyland:
*Issue #, if available:*

*Description of changes:*
We get this error `missing [bootstrap-token] phase(s) needed by "show-join-command" phase` when attempting to create a 1.26 BR cluster because this [PR](https://github.com/kubernetes/kubernetes/pull/111512) in K8s v1.26 moved show-join-command as a new separate phase at the end of kubeadm init. It also mentioned that executing only this phase on demand will throw an error because the phase needs dependencies such as bootstrap tokens to be pre-populated. This matched the error what we observed.

This PR should fix the error. 


projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
```diff
@@ -125,7 +125,7 @@ func controlPlaneInit() error {
 	}
 
 	// finish kubeadm
-	cmd = exec.Command(kubeadmBinary, utils.LogVerbosity, "init", "--skip-phases", "preflight,kubelet-start,certs,kubeconfig,bootstrap-token,control-plane,etcd",
+	cmd = exec.Command(kubeadmBinary, utils.LogVerbosity, "init", "--skip-phases", "preflight,kubelet-start,certs,kubeconfig,bootstrap-token,control-plane,etcd,show-join-command",
```
jaxesn, line 128: does this work in older versions of kube? i assume so since it just a list, id assume kubeadm is smart enough to skip over ones it doesnt know?

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
```diff
@@ -124,9 +126,29 @@ func controlPlaneInit() error {
...
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "Error running command: %v, Output: %s\n", cmd, string(out))
+	}
+	kubeadmVersion, err := versionutil.ParseSemantic(string(out))
+	if err != nil {
+		return errors.Wrapf(err, "%q is not a valid kubeadm version", string(out))
+	}
+
+	compare, err := kubeadmVersion.Compare("1.26")
```
jaxesn, line 139: We probably want this to be >= 1.26.

In general, this is probably fine to get the version from the kubeadm binary, but i suspect the kube version would also be in the kubeadm config. Not sure if its necessarily a better spot to get it from, but maybe we are already getting this somewhere else in the code? thoughts @vignesh-goutham

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
```diff
@@ -124,9 +126,29 @@ func controlPlaneInit() error {
 		return errors.Wrap(err, "Error waiting for kubelet to come up")
 	}
 
+	cmd = exec.Command(kubeadmBinary, utils.LogVerbosity, "version", "-o=short")
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "Error running command: %v, Output: %s\n", cmd, string(out))
```
jaxesn, line 132: Honestly i could be wrong here and im not sure what we do elsewhere during this bootstrap process, but failing during bootstrap is really bad and potentially hard to diagnosis. I wonder if its better in cases like, if we cant get or compare the version, we should just pick a default setup of skips to do, vs failing the entire process?

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
taneyland: Yes, if it's less than 1.26, we don't skip the new phase (bc it didn't exist in the older versions), but if 1.26+, we skip the new phase.

But yes, open to other thoughts if we dont want to use `kubeadm version`

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
taneyland: We error out during this process throughout this file. I do agree that something like this comparison might not justify failing the entire process, but it will fail later in the bootstrap process if the we don't skip the correct phases based on the kube version

vignesh-goutham: Yea we do fail for commands like this throughout, but we print those logs to journalctl. So if this were to error out, we can still get the logs.

vignesh-goutham (approved):

PR #2018, opened 2023-03-20
Bumping packages to v0.3.7

junshun:
This version will include fix to not run credential package on BR 1.25 nodes which was causing kubelet crashes on those nodes.


TerryHowe (approved):
/lgtm
/approve

PR #2086, opened 2023-04-13
Adds builds for containerd and runc

jaxesn:
*Issue #, if available:*

*Description of changes:*

During our image builder process which users run on their own, containerd and runc are pulled from upstream's releases.  To better match the rest of our dependencies strategy, we are looking at building these two along with everything else and using these builds instead of the upstream builds.

As a follow to this, I am going to modify the containerd build to package runc along with it in the same tgz to match upstreams payload and make it easier to use in the image-builder scripts directly.


projects/containerd/containerd/README.md
```diff
@@ -0,0 +1,21 @@
+## **Source Controller**
```
abhay-krishna, line 1:
```suggestion
## **containerd**
```

abhay-krishna (approved):
/approve

PR #2107, opened 2023-04-20
Patch haproxy dynmanic config via kind cli as well

jaxesn:
*Issue #, if available:*

*Description of changes:*

My last patch was technically incomplete, but for eks-a this is not actually necessary since our use of the haproxy loadbalancer is only thru CAPD which does not use this particular kind code path. Patching anyway to be consistent.


abhay-krishna:
/lgtm
/approve
/override imagebuilder-1-25-presubmit imagebuilder-1-24-presubmit

PR #2110, opened 2023-04-21
Fix Autoscaler Generate Attribution

jonathanmeier5:
Required a few changes:
- Marking autoscaler/eks-anywhere-patched as a dependency of remove-cloud-providers so the project directory is always downloaded and checked out
- Marking remove-cloud-providers as a dependency of all-attribution


projects/kubernetes/autoscaler/Makefile
```diff
@@ -35,9 +35,10 @@ include $(BASE_DIRECTORY)/Common.mk
 
 
 $(GO_MOD_DOWNLOAD_TARGETS): remove-cloud-providers
+attributions: remove-cloud-providers
 
 .PHONY: remove-cloud-providers
-remove-cloud-providers:
+remove-cloud-providers: autoscaler/eks-anywhere-patched
```
jaxesn, line 41: I think there is a var for this, like GIT_PATCH_TARGET

jaxesn (approved):

abhay-krishna (approved):
/approve

PR #2116, opened 2023-04-25
Adding a patch to limit rufio bmclib drivers

pokearu:
## Description
bmc-lib tries to establish a connection to check compatible protocols (redfish, ipmi etc) for all supported implementations. These changes modify Rufio to only use `Redfish` and `IPMI` protocols for all hardware.

## Why is this needed
Rufio opens a new connection via bmc-lib for every reconcile loop. In the latest version of bmc-lib support for IntelAMT protocol was added. It was noticed that on older hardware when users try to run Rufio, IntelAMT takes too long to establish a connection and hence blocks the reconcile thread. This was causing timeouts of Rufio Tasks when it exceeded 3 mins to hit the connection timeout. 

Fixes:
Timeout issues on older hardware due to Rufio reconcile threads getting blocked when bmc-lib provider open connection calls hang and timeout. 

## How Has This Been Tested?
`make docker-build` and tested the controller.


mitalipaygude (approved):

abhinavmpandey08 (approved):
/woof
/lgtm

PR #2156, opened 2023-05-11
Make image-builder check out a specific commit when running builds

### LINKED ISSUE #2136: image-builder should checkout a specific release commit before running builds
Currently the working of image-builder for EKS-A customers involves cloning the aws/eks-anywhere-build-tooling repo from GitHub and kicking off the build from the code in main. This introduces the possibility of breaking changes in main affecting customers. For example, there were two separate instances of customers reporting image build failures (#2125, aws/eks-anywhere#5723) due to containerd URL being set to `null`. This is because we switched to using our own builds of containerd from the bundle manifest, and that works for image builds in CI because the dev bundle contains a containerd URI field, but the prod release manifest doesn't have it. This was temporarily fixed in #2134 but we should put in a robust fix by making the image-builder workflow checkout a specific commit instead of just running it from main.

abhay-krishna:
Fixes #2136.


projects/aws/image-builder/builder/utils.go
```diff
@@ -94,7 +104,24 @@ func execCommand(cmd *exec.Cmd) (string, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("failed to run command: %v", err)
+		return commandOutputStr, fmt.Errorf("failed to run command: %v", err)
 	}
 	return commandOutputStr, nil
 }
+
+func getGitCommitFromBundle(repoPath string) (string, error) {
+	log.Println("Getting git commit from bundle")
+	loadBundleManifestCommandSequence := fmt.Sprintf("chmod +x %s/build/lib/eksa_releases.sh && source %s/build/lib/eksa_releases.sh && build::eksa_releases::load_bundle_manifest", repoPath, repoPath)
```
jaxesn, line 114: This whole thing feels a bit strange. Feels like would be better to reimplement this in go, but I think I get why you did this and I suspect we do this kind of thing already since this image builder is really just a wrapper around our build tooling.

The chmod tho, can't we do that just in git and not have to do this everytime?

@vignesh-goutham please review this

abhay-krishna: Ah yeah I was gonna remove that. I put it in code to test locally because it clones the repo everytime and the one on Github doesn't have it.

TBH I shouldn't even need `chmod` in the first place,  but the `source` command run through Go code kept complaining that the file wasn't present even though it was. When run on the terminal, it works fine as it's a shell builtin.

vignesh-goutham (approved):

PR #2158, opened 2023-05-11
Ensure source AMI architecture is AMD64

abhay-krishna:
Most public AMIs matching the filter `amzn2-ami-kernel-5.*-hvm-2*` come in both AMD64 and ARM64 variants, where the build chooses the AMD64 variant. However filtering for the latest AL2 Kernel 5.10 AMI returned only one source AMI (`ami-0ae285eae0f72f673`) which was ARM64, and caused the build to fail with the message
```
Error launching source instance: InvalidParameterValue: The architecture 'x86_64' of the specified instance
type does not match the architecture 'arm64' of the specified AMI. Specify an instance type and an AMI that
have matching architectures, and try again. You can use 'describe-instance-types' or 'describe-images' to
discover the architecture of the instance type or AMI.
```
So we need to filter for an AMD64 source AMI even if it means we get a slightly older AMI.


pokearu (approved):
/woof
/approve
/lgtm

PR #2210, opened 2023-06-06
bumps flux to ga rc.5

jaxesn:
*Issue #, if available:*

*Description of changes:*

Bumps to the latest ga rc of flux.  The main thing to note here is the update of the helm controller which calls out the updating of kustomize to v5 having breaking changes: https://github.com/fluxcd/helm-controller/blob/v0.34.1/CHANGELOG.md

@jiayiwang7 please confirm if we need to make cli changes to accommodate.  Id also like to back port this to release-0.16 since we shipped rc releases, ideally they go GA some time soon and we backport that update as well.


projects/fluxcd/flux2/patches/0001-Replace-github.com-homeport-dyff-with-commit-to-work.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/go.mod b/go.mod
+index f65edc85..68eaff7e 100644
+--- a/go.mod
++++ b/go.mod
+@@ -217,3 +217,5 @@ require (
+ 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
+ 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+ )
++
++replace github.com/homeport/dyff => github.com/homeport/dyff v1.5.7-0.20230415202303-35afb2435b21
```
jaxesn, line 21: this newest version of dyff appears to have been republished which gets the go sum hashes out of date.  We hit this every now and then and the "fix" is to point to a commit either before or right after the release tag. This commit is the commit before the tag.

projects/kubernetes-sigs/image-builder/packer/ami/additional_files/bootstrap.sh
```diff
@@ -16,15 +16,20 @@ echo "127.0.0.1   $(hostname)" >>/etc/hosts
 NETWORK_CONFIG_PATH=/tmp/network.yaml
 NETPLAN_CONFIG_PATH=/etc/netplan/config.yaml
 
+# dynamically get default network interface name using mac address
```
jiayiwang7, line 19: u may need to rebase from main

jaxesn: oh yeah wth is this!?

jaxesn:
nm we dont need hold
/unhold

PR #2256, opened 2023-06-23
Fix metadata service queries treating 404s as OK

chrisdoherty4:
Fixes https://github.com/aws/eks-anywhere/issues/5564

When the Hegel metadata service doesn't have a Hardware record it returns 404. Our writefile Tinkerbell action treats the 404 as OK resulting in invalid metadata being written to disk and ultimately consumed by kubelet on launch, causing it to fail.

This change treats non-200 responses as errors ensuring we move to query any additional metadata endpoints.

### DIFF
--- a/projects/tinkerbell/hub/patches/0010-Handle-non-200-status-codes-from-metadata-service.patch
+++ b/projects/tinkerbell/hub/patches/0010-Handle-non-200-status-codes-from-metadata-service.patch
@@ -1,0 +1,28 @@
+From 9935e57fc9fb02b2af061f1de7ea9976c1b6ab17 Mon Sep 17 00:00:00 2001
+From: Chris Doherty <chris.doherty4@gmail.com>
+Date: Fri, 23 Jun 2023 09:18:26 -0500
+Subject: [PATCH] Handle non-200 status codes from metadata service
+
+---
+ actions/writefile/v1/main.go | 5 +++++
+ 1 file changed, 5 insertions(+)
+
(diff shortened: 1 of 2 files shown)


jacobweinstock (approved):

projects/tinkerbell/hub/patches/0010-Handle-non-200-status-codes-from-metadata-service.patch
```diff
@@ -0,0 +1,28 @@
...
+index db393fe..954a45e 100644
+--- a/actions/writefile/v1/main.go
++++ b/actions/writefile/v1/main.go
+@@ -205,6 +205,11 @@ func main() {
+ 			}
+ 			defer resp.Body.Close()
+ 
++			// Ensure non-200 responses are considered an error and move to the next metadata server.
++			if resp.StatusCode != http.StatusOK {
++				continue
```
cxbrowne1207, line 20: How is it that we can use `continue` here? Is this within some kind of loop? I'm not too familiar with the syntax in patches.

jacobweinstock: yeah, exactly. its in a for loop. And i only know that because im familiar with the code base. Unfortunately, with the way we do patches you're not able to see this context in the PR. And to see the proper context, you'd have to apply all our patches first. I.e. clone this repo and run then `make clone-repo`, `make checkout-repo`, and `make patch-repo`.  :(

PR #2219, opened 2023-06-08
adds support for ubuntu 2204 in image-builder

jaxesn:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2264, opened 2023-06-27
Bump image-builder to v0.1.16

adiantum:
Description of changes:
 - bump image-builder to v0.1.16
 - update patches
 - delete patch 0016 because already in upstream: https://github.com/kubernetes-sigs/image-builder/commit/2a4bdfb17c12392ef3d17506c93c2a95689587b6?diff=unified

Nutanix image tested successfully. 


jaxesn (commented):
Awesome! Thanks for the contribution!

I started this work in the context of adding support for ubuntu 22.04 in my [PR](https://github.com/aws/eks-anywhere-build-tooling/pull/2219) here. Im happy to approve and merge this one tho so we can get it in sooner.  

The one thing I changed in my attempt was the two patches where we add a few vars to the ansible list. This patch always conflicts and requires us to update it.  I was thinking we could move the 3 or 4 additional vars we add to the end of the list instead of in the middle where they are now to try and avoid the diff everytime. Any thoughts on this?  Would you mind updating yours to do the same, or feel free to push back.

adiantum:
> The one thing I changed in my attempt was the two patches where we add a few vars to the ansible list. This patch always conflicts and requires us to update it. I was thinking we could move the 3 or 4 additional vars we add to the end of the list instead of in the middle where they are now to try and avoid the diff everytime. Any thoughts on this? Would you mind updating yours to do the same, or feel free to push back.

I think that's great idea! I believe it should help to maintain there patches in future. I'll update the PR based on your work.

adiantum:
@jaxesn I made the changes in ansible vars patches. Please take a look.

jaxesn (approved):
Awesome thanks!

/approve

PR #2269, opened 2023-06-27
Bump CAPI and cert-manager versions

abhay-krishna:
Bumping versions for CAPI and cert-manager.


abhay-krishna:
/override eks-anywhere-cli-tools-tooling-presubmit
expected failure

g-gaston (commented):
/lgtm

build/lib/simple_create_tarballs.sh
```diff
@@ -25,8 +25,8 @@ BINARY_PLATFORMS="$5"
 TAR_PATH="$6"
 GIT_HASH="$7"
 
-LICENSES_PATH="$OUTPUT_DIR/LICENSES"
-ATTRIBUTION_PATH="$OUTPUT_DIR/ATTRIBUTION.txt"
+LICENSES_PATHS=($(find $OUTPUT_DIR -type d -name "LICENSES"))
```
jaxesn, line 28: oh dang im surprised this wasnt already handled. i guess we just didnt notice for projects with multiple attr files they werent getting put in the tarball?  or maybe this is the first one that produces a tarball and multiple attr files?

abhay-krishna: Yes even i was surprised. There are others like the tinkerbell actions that produce multiple attribution files but we don't have tarball for those

abhay-krishna:
/override kind-tooling-presubmit
This issue won't be seen in Codebuild

PR #2274, opened 2023-06-29
Support installing extra Debian packages and RPMs during image builds

abhay-krishna:
Users may want to install additional packages during build time on their EKS-A nodes. Upstream image-builder supports this, so adding support in our image-builder to pass it through to upstream.


jaxesn (approved):
lets plan on updating the docs to call this out

PR #2306, opened 2023-07-10
disables udp offload on redhat for vwmare

jaxesn:
*Issue #, if available:*

*Description of changes:*

When updating to cilium 1.12 we ran into the udp overload issue on vmware with Redhat 8. We have seen this problem off an on, seems to depend on the specific version of vpshere, with BR for a while due to their use are pretty new kernels.  Recently they implemented a [solution](https://github.com/bottlerocket-os/bottlerocket/pull/2850) by disabling udp offloading.  There is also potentially some official [guidance](https://github.com/cilium/cilium/issues/21801) coming from the cilium community on this issue for vmware environments. 

This patches our build of image builder to add a similar systemd unit to the Redhat build for vmware.  @taneyland has tested this manually.

### DIFF
--- a/projects/kubernetes-sigs/image-builder/patches/0007-Support-crictl-validation-from-input-checksum.patch
+++ b/projects/kubernetes-sigs/image-builder/patches/0007-Support-crictl-validation-from-input-checksum.patch
@@ -1,7 +1,7 @@
 From eb0ec6499bcf05e4683215c6bbcee1a74cbecdcc Mon Sep 17 00:00:00 2001
 From: Vignesh Goutham Ganesh <vgg@amazon.com>
 Date: Fri, 2 Sep 2022 14:32:21 -0700
-Subject: [PATCH 08/17] Support crictl validation from input checksum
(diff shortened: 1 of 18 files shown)


projects/kubernetes-sigs/image-builder/patches/0007-Support-crictl-validation-from-input-checksum.patch
```diff
@@ -1,7 +1,7 @@
 From eb0ec6499bcf05e4683215c6bbcee1a74cbecdcc Mon Sep 17 00:00:00 2001
 From: Vignesh Goutham Ganesh <vgg@amazon.com>
 Date: Fri, 2 Sep 2022 14:32:21 -0700
-Subject: [PATCH 08/17] Support crictl validation from input checksum
```
abhay-krishna, line 4: This should be 7, and the numbers from 7 to 13 need to be fixed.

abhay-krishna (approved):

PR #2307, opened 2023-07-10
Fix getting git commit from bundle cmd

d8660091:
*Issue #, if available:* `build --os ubuntu --hypervisor baremetal --release-channel 1-24` cannot run because it fails to fetch git commit

*Description of changes:* fix the command to fetch git commit

### DIFF
--- a/build/lib/eksa_releases.sh
+++ b/build/lib/eksa_releases.sh
@@ -21,7 +21,7 @@ function build::eksa_releases::load_bundle_manifest() {
   local -r dev_release=${1-false}
   local -r latest=${2-latest}
   local -r echo=${3-true}
-  oldopt=$-
+  oldopt="$(set +o)"
   set +o nounset
   set +x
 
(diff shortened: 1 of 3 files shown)


build/lib/eksa_releases.sh
```diff
@@ -21,7 +21,7 @@ function build::eksa_releases::load_bundle_manifest() {
   local -r dev_release=${1-false}
   local -r latest=${2-latest}
   local -r echo=${3-true}
-  oldopt=$-
+  oldopt="$(set +o)"
```
jaxesn, line 24: is this due to a bash version diff or something?  this is fine for now, but as a followup, lets remove the bash call to this altogether and just "curl" in golang directly. cc @abhay-krishna

d8660091: Since we invoke the function with "bash -c", there is an invalid opt "c" in $-. There is an error when you run "set -$oldopt", because you cannot run "set -c"

abhay-krishna: +1. I am okay with this as a temporary solution, but we can change it to a native URL read in Go

jaxesn (approved):

PR #2314, opened 2023-07-12
Add Go version to upstream projects tracker

abhay-krishna:
This change moves the Go version definition from the project Makefile to a separate tracking file. The Go version for a project has also been added to the upstream projects tracker file.


projects/apache/cloudstack-cloudmonkey/Makefile
```diff
@@ -1,6 +1,6 @@
 BASE_DIRECTORY:=$(shell git rev-parse --show-toplevel)
 GIT_TAG:=$(shell cat GIT_TAG)
-GOLANG_VERSION?="1.20"
+GOLANG_VERSION?=$(shell cat GOLANG_VERSION)
```
jaxesn, line 3: can you replace all of these with `:=` instead of `?=`

jaxesn (changes requested):
I like this!

We didnt have to split the versions into a separate file, we could have called make to get the golang version like we do in our buildspec generation scripts.  This is a pattern we have followed tho for release branch'd builds so im fine with the change.

abhay-krishna:
> I like this!
> 
> We didnt have to split the versions into a separate file, we could have called make to get the golang version like we do in our buildspec generation scripts.  This is a pattern we have followed tho for release branch'd builds so im fine with the change.

True, the reason I split them even for non release branched projects is to make the automation for Git tags/go version updates cleaner. Felt it would be better to update the single tracking file instead of invoking sed on the Makefile.

jaxesn (approved):

PR #2358, opened 2023-07-27
Update capi etcd patch to avoid NodeHealthy false

g-gaston:
*Issue #, if available:*
Fixes https://github.com/aws/eks-anywhere/issues/5893

*Description of changes:*
For an easier review: https://github.com/abhay-krishna/cluster-api/pull/24/commits/99a91f2e7219d11f900a8d2509d49e0294fbb039

There is no impact in the execution flow. It's mostly to avoid confusion when looking at the machine status.


projects/kubernetes-sigs/cluster-api/patches/0001-Adding-capi-support-for-Bottlerocket.patch
```diff
@@ -1,4 +1,4 @@
-From c0d89ec2a34390253ed684c6f375e7c98309db66 Mon Sep 17 00:00:00 2001
+From 10dd70d5cf1eaf47cd2060d3e83847058072af59 Mon Sep 17 00:00:00 2001
 From: Vignesh Goutham Ganesh <vgg@amazon.com>
 Date: Fri, 11 Jun 2021 10:43:09 -0700
 Subject: [PATCH 01/34] Adding capi support for Bottlerocket
```
jiayiwang7, line 4:
```suggestion
Subject: [PATCH 01/35] Adding capi support for Bottlerocket
```

same for all

g-gaston: why does it need to change to 35? I only see 34 patches. Also, this is auto-generated.

PR #2361, opened 2023-07-27
Embed EKS-A version in image-builder with support for overrides

abhay-krishna:
This PR adds the logic to embed the EKS-A version in production image-builder CLI binaries during build time to make it easy for customers to understand what version of EKS-A the built images will correspond to. An `eksa-release` flag is also added for customers to be able to override the EKS-A version for which images are to be built. Based on the version passed or embedded, the corresponding bundle will be used to get the artifacts and git commit to checkout.


projects/aws/image-builder/builder/utils.go
```diff
@@ -109,19 +112,70 @@ func execCommand(cmd *exec.Cmd) (string, error) {
...
+	if err = yaml.Unmarshal(releasesManifestContents, releases); err != nil {
+		return "", fmt.Errorf("failed to unmarshal release manifest from [%s]: %v", eksAReleasesManifestURL, err)
+	}
+
+	var eksAReleaseVersion, bundleManifestUrl string
+	var foundRelease bool
+	if bo.EKSAReleaseVersion != "" {
+		eksAReleaseVersion = bo.EKSAReleaseVersion
+	} else {
+		eksAReleaseVersion = releases.Spec.LatestVersion
```
jaxesn, line 132: lets put out a warning to the user here to make it clear we are using the latest

`No eks-a-release specified, using the latest version of <>`

jaxesn (changes requested):
looks reasonable, can you try and test it against 0.16/0.15 and reelase-0.17? mainly confirming the git hashes and at least for 0.16 make sure the make targets work or whatever, somehow.

projects/aws/image-builder/builder/utils.go
abhay-krishna: Done

projects/aws/image-builder/builder/utils.go
```diff
@@ -109,19 +112,76 @@ func execCommand(cmd *exec.Cmd) (string, error) {
...
+
+func readFileFromURL(url string) ([]byte, error) {
+	request, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed creating http GET request for downloading file: %v", err)
+	}
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSHandshakeTimeout = 60 * time.Second
+	client := &http.Client{
```
jaxesn, line 172: do we need to explictly set the http_proxy stuff when we do this?

projects/aws/image-builder/Makefile
```diff
@@ -16,17 +19,25 @@ REPO_NO_CLONE=true
...
 
-include $(BASE_DIRECTORY)/Common.mk
 
+include $(BASE_DIRECTORY)/Common.mk
 
 $(REPO):
 	@mkdir $@
+	if [[ "$(CODEBUILD_BUILD_ID)" =~ "aws-staging-bundle-build" ]]; then \
+		source $(BUILD_LIB)/common.sh && retry git clone --depth 1 --filter=blob:none --sparse -b $(BRANCH_NAME) $(EKS_ANYWHERE_CLONE_URL) $(EKS_ANYWHERE_REPO); \
+		git -C $(EKS_ANYWHERE_REPO) sparse-checkout set $(EKS_ANYWHERE_REPO_SPARSE_CHECKOUT) --cone --skip-checks; \
```
jaxesn, line 35: is `--cone` right?

overall this looks scary... where is BRANCH_NAME coming from?  and will the CLI_MX_VERSION file always be the right verison?

projects/aws/image-builder/builder/utils.go
abhay-krishna: not necessary

jaxesn (approved):

PR #2374, opened 2023-08-10
Provide default value for EKSA_USE_DEV_RELEASE variable

abhay-krishna:
* Providing default value for EKSA_USE_DEV_RELEASE variable so that we don't get unbound variable errors.
* Pinning packer-plugin-nutanix to version v0.7.1 so that we avoid including [this change](https://github.com/nutanix-cloud-native/packer-plugin-nutanix/pull/142) that breaks presubmits due to empty configs.


vignesh-goutham (approved):

abhay-krishna (commented):
/retest

vignesh-goutham (approved):

PR #2405, opened 2023-08-19
disable cgroupns in kind/capd

jaxesn:
*Issue #, if available:*

*Description of changes:*

Not totally sure why but using cgroupns=private on AL2 nodes, this did not seem to be an issue on Ubuntu 20.04 nodes which is also cgroup v1, seems to cause issues creating the containers instead the kind node image.

Kind 0.20 does call out that using cgroupns=default with cgroupv1 is deprecated and will eventually be removed.  The code to handle this case has been in the entrypoint.sh script for quite some time and we *can* continue to maintain this in our patches if necessary.  

Newer distro, Ubuntu 22.04 and AL23, use cgroup v2 which are not affected by this change since docker already defaults to cgroupns=private for cgroup v2 hosts.

I confirmed this issue exist on an AL2 using the upstream kind binary and upstream kind node image as well.  Using this patched version of kind, I was able to create a cluster with both the upstream image and the eks-a al23 based node image.


jaxesn:
Merging to get a build to test, will revert if needed.

PR #2406, opened 2023-08-20
revert runc and containerd in kind image

jaxesn:
*Issue #, if available:*

*Description of changes:*

I tested on an al2 node using a combo of upstream and our custom builds of kind + node image. I *think* the updated runc appears to be the problem, even tho the 0.20 release of kind added a fix to workaround the misc controller issue.  

The main problem comes in when installing cilium.  Creating a bootstrap cluster with kindnetd does not appear to have the same issues. I tested the eksa 1.12 clium as well as the upstream 1.12 and 1.13 cilium, all of which show the issue.  Upstream 1.14 cilium does not have the same problem.

I am not completely convinced there is nothing some other way to work around this, but this seems to the simplest for now.


jaxesn:
Disk space issue in presubmit.. I'll fix it a follow-up

PR #2412, opened 2023-08-21
do not lock containerd version

jaxesn:
*Issue #, if available:*

*Description of changes:*

Follow up to https://github.com/aws/eks-anywhere-build-tooling/pull/2406

Tested this on Al2 and the newer containerd version does not seem to be a problem.  This makes sense because the issue we looks very much like the runc 1.1.6/1.1.7 misc controller issue seen in kind/kubernetes.


abhay-krishna (approved):
/lgtm
/approve

PR #2417, opened 2023-08-21
adds retries and timeout to packer image-builder

jaxesn:
*Issue #, if available:*

*Description of changes:*

This adds retries to initial ansible tasks as well as changes the ssh_timeout, which we do see run forever sometimes, to 20 mins which matches the nutanix config.  I *think* this may (automatically?) retry in the cases where we see timeouts and failures today. It may just make the failure quicker, in which case we can add a followup PR to add retries.  

Want to try this first and see what happens.


abhay-krishna (approved):
/approve

PR #2434, opened 2023-08-25
Fall back to previous release branch section for common binaries

abhay-krishna:
Image builds parse the (dev) bundle manifest to retrieve tarball URLs for binaries such as `crictl`, `etcdadm` and `containerd`. The bundle is divided into several sections, one for each EKS-D release branch (Kubernetes version) and the scripts pass in the release branch to get the URL from that section. When building images for a new release branch for the first time, the section for that release branch would not have been added to the bundle yet, so the URLs will be returned as empty strings by the `yq` query. So we fall back to the previous release branch section to get the URLs. These artifacts for these projects are common across Kubernetes versions so we can use the artifact URL from any other release branch.

### DIFF
--- a/build/lib/eksa_releases.sh
+++ b/build/lib/eksa_releases.sh
@@ -15,6 +15,7 @@
 
 # Key for bundles manifest is environment/latest
 # example key = "dev/release-1.20"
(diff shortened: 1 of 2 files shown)


jaxesn (approved):
/approve

PR #2450, opened 2023-08-29
Bump capi to v1.4.6 and cert-manager to v1.12.3

g-gaston:


jaxesn:
want to backport this?

g-gaston:
> want to backport this?

@jaxesn I wasn't planning on doing it but we could.

However, I don't want to do the rebase again, it was a pain (release-0.17 would require a new rebase of the patches since they are different that in main). So if we feel confident with the current patches in main (we would need to check the e2e signal for the last few days), then I would suggest just copying all the patches from main to avoid the extra work.

I don't see anything in the changelog that we **need** in v0.17.1.

Issue #2372, opened 2023-08-10
Passing packer variables to specify custom overrides does not seem to work


vasyl-purchel:
**What happened**:

Command in use to build Ubuntu 22.04 based EKSA BM image:

```shell
image-builder-dev4 build --os ubuntu --os-version 22.04 --hypervisor baremetal --release-channel 1-27 -v 3 --force --baremetal-config /home/image-builder/packer_vars.json
```

I have `/home/image-builder/packer_vars.json` file with overrides to use a custom Ansible role through image provisioning.
When I attempt to pass it with `--baremetal-config /home/image-builder/packer_vars.json` this path is nowhere to be seen in end command under `PACKER_VAR_FILES` environment variable.

Also I've noticed following warning during execution:

```
Makefile:46: warning: undefined variable 'PACKER__VAR_FILES'
```

which looks like comes from:

```makefile
PACKER_TYPE_VAR_FILES?=$(PACKER_$(call TO_UPPER,$(IMAGE_FORMAT))_VAR_FILES)
```

where `IMAGE_FORMAT` is empty instead of `raw`.

I've also attempted to modify `commandEnvVars` for Ubuntu build with no impact:

```diff
                 case Ubuntu:
                         buildCommand = fmt.Sprintf("make -C %s local-build-raw-ubuntu-%s", imageBuilderProjectPath, b.OsVersion)
+                        commandEnvVars = append(commandEnvVars, fmt.Sprintf("PACKER__VAR_FILES='/home/image-builder/packer_vars.json'"), fmt.Sprintf("PACKER_RAW_VAR_FILES='/home/image-builder/packer_vars.json'")
                 case RedHat:
```

End result packer command executed:

```
[...]

Issue #2301, opened 2023-07-07
Feature: Allow users to add custom packages to images built with image-builder


chrisdoherty4:
As a user, I want to add custom packages to images built with image builder (Ubuntu in particular) during the build process so I can streamline my provisioning process and avoid manually adding packages after the fact.

Packages may require additional configuration steps.

vignesh-goutham:
The built-in apt ansible package cannot pull deb packages from an http endpoint. On the contrary yum ansilbe package can pull down and install rpms from an http endpoint in addition to pulling them from the repo. Image-builder should patch and maintain a way for users to pull down and install a custom deb package for Ubuntu image builds.

chrisdoherty4:
Users should be able to execute custom installers.

PR #2474, opened 2023-09-07
Adds pre/post boot and node roles

vasyl-purchel:
*Issues addressed:*

* #2301 
* #2372

*Description of changes:*
As per https://github.com/kubernetes-sigs/image-builder/issues/894 issue upstream project opened to use custom Ansible roles pre/post boot of the node and also pre/post node provisioning.

This opens doors to customize image built for custom requirements, for example if deployment needs to happen in offline environment that has no access to the internet and would require to update system with required changes to look for packages on internal proxies instead.

Also allows to provide extra updates to the system before Kubernetes is installed along with extra tweaks after.

### DIFF
--- a/projects/aws/image-builder/builder/types.go
+++ b/projects/aws/image-builder/builder/types.go
@@ -71,6 +71,7 @@ type VsphereConfig struct {
 	RhelConfig
 	ProxyConfig
 	ExtraPackagesConfig
+	ExtraOverridesConfig
 }
 
 type BaremetalConfig struct {
@@ -79,6 +80,7 @@ type BaremetalConfig struct {
 	RhelConfig
 	ProxyConfig
 	ExtraPackagesConfig
+	ExtraOverridesConfig
 }
 
 type CloudstackConfig struct {
@@ -87,6 +89,7 @@ type CloudstackConfig struct {
 	RhelConfig
 	ProxyConfig
 	ExtraPackagesConfig
+	ExtraOverridesConfig
 }
 
 type IsoConfig struct {
@@ -114,6 +117,7 @@ type NutanixConfig struct {
 	NutanixSubnetName string `json:"nutanix_subnet_name"`
 	ProxyConfig
 	ExtraPackagesConfig
+	ExtraOverridesConfig
 }
 
 type AMIConfig struct {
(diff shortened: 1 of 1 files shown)


vasyl-purchel:
PR verified:

```shell
image-builder@eksa_test:~$ cat /home/image-builder/packer_vars.json
{
        "disable_public_repos": "true",
        "node_custom_roles_pre": "wd.eksa.pre_node"
}
image-builder@eksa_test:~$ cat ~/.ansible/roles/wd.eksa.pre_node/tasks/main.yml 
- name: Do custom stuff
  apt:
    name:
      - jq
      - ansible
      - make
    update_cache: yes
    state: latest
image-builder@eksa_test:~$ export EKSA_USE_DEV_RELEASE=true
image-builder@eksa_test:~$ image-builder-pr build --os ubuntu --os-version 22.04 --hypervisor baremetal --release-channel 1-27 -v 3 --force --baremetal-config /home/image-builder/packer_vars.json
...
```

and while it's running we can see that overrides are passed over to packer:

```shell
$ cat /home/image-builder/eks-anywhere-build-tooling/projects/kubernetes-sigs/image-builder/packer/config/baremetal.json
{"files":null,"rhel_username":"","rhel_password":"","node_custom_roles_pre":"wd.eksa.pre_node","disable_public_repos":"true"}
```

> Note: this example is not complete since when we disable public repos we need to supply private repos as well and for ansible roles I'm using only one role while value can be space separated multiple roles

vignesh-goutham:
Thanks for this! 

One question I had was, can the pre/post role take in a full path for the ansible role, and if user doesn't provide the full path, does it expect the role to be present in a certain relative path?

Also could you add this overrides to cloudstack, vsphere, nutanix and ami builder types?

vasyl-purchel:
details on what these overrides can take are very well described here: https://image-builder.sigs.k8s.io/capi/capi.html#customization
if ansible is not configured to look in specific location it can't take full path to it, but if you put roles into ansible configured location then it can use those

in my case I've used `~/.ansible/` where I believe ansible-galaxy would install collections and they were found out.

as for adding those to other hypervisors can do for sure but won't be able to verify if it works in any way as I don't have those configured

vasyl-purchel:
@vignesh-goutham , I've updated PR with requested addition to other hypervisor types along with rebasing the change on top of latest main branch.

PR #2509, opened 2023-09-21
cherrypick: Adds pre/post boot and node roles (#2474)

jaxesn:
As per https://github.com/kubernetes-sigs/image-builder/issues/894 issue upstream project opened to use custom Ansible roles pre/post boot of the node and also pre/post node provisioning.

This opens doors to customize image built for custom requirements, for example if deployment needs to happen in offline environment that has no access to the internet and would require to update system with required changes to look for packages on internal proxies instead.

Also allows to provide extra updates to the system before Kubernetes is installed along with extra tweaks after.

*Issue #, if available:*

*Description of changes:*


vignesh-goutham (approved):
/approve

PR #2522, opened 2023-10-02
improves use of gnu variants on macosx

jaxesn:
*Issue #, if available:*

*Description of changes:*

We had a common "pattern" for certain tools, such as sed + tar +date, where on mac we would try and find the gnu build, which is installable via brew.  This introduces a helper and refactors our usage of this pattern so its consistent throughout our scripts.

This will also make it a bit more easier to track which brew formulas someone should install if building locally. I plan to update the building-locally doc in a future PR.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -302,7 +302,7 @@ ADD_TRAILING_CHAR=$(if $(1),$(1)$(2),)
 # check if pass variable has length of 1
 IS_ONE_WORD=$(if $(filter 1,$(words $(1))),true,false)
 
-SED_CMD=$(shell if [ "$$(uname -s)" = "Darwin" ] && command -v gsed &> /dev/null; then echo gsed; else echo sed; fi)
+SED_CMD=$(shell source $(BUILD_LIB)/common.sh && build::common::gnu_variant_on_mac sed)
 
 ####################################################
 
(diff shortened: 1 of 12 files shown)


Common.mk
```diff
@@ -842,7 +842,8 @@ release: $(RELEASE_TARGETS)
 # Iterate over release branch versions, avoiding branches explicitly marked as skipped
 .PHONY: %/release-branches/all
 %/release-branches/all:
-	@for version in $(SUPPORTED_K8S_VERSIONS) ; do \
+	@set -e; \
```
jaxesn, line 845: We actually have this configured at the top, however the version of make on mac does not support the global flag.  For this specific target, its annoying not to fail on error so adding it specifically.

build/lib/common.sh
```diff
@@ -129,7 +149,7 @@ function build::gather_licenses() {
   # data about each dependency to generate the amazon approved attribution.txt files
   # go-deps is needed for module versions
   # go-licenses are all the dependencies found from the module(s) that were passed in via patterns
-  build::common::echo_and_run go list -deps=true -json ./... | jq -s ''  > "${outputdir}/attribution/go-deps.json"
+  build::common::echo_and_run go list -deps=true -json ./... | jq -s '.'  > "${outputdir}/attribution/go-deps.json"
```
jaxesn, line 152: i updated jq on my mac and the newer version expects `.`, i validated this works the same on the old jq as well.

PR #2546, opened 2023-10-12
refactors logging to use annonation style

jaxesn:
*Issue #, if available:*

*Description of changes:*

While working on improving the docker usage via the build targets, basically transparently using docker where necessary instead of relying on the eng to run the `in-docker` target, I found an interesting (probably hacky) approach for adding target variables via a secondarily expanded prereq.  In that case I am using this approach to dynamically add the docker target as a prereq to the real target and then overriding the shell to `true` so that the on host command do not even run, since they ran in the docker container.  This seems to work pretty well but I have a problem regarding how make outputs the commands before it runs them.

While trying to solve the problem of outputting these commands, even tho they were not actually going to run, I was playing with this idea of using a bash script as the shell we could better control this output.  This PR introduces an approach where instead of calling log_start and log_end in the recipe a secondarily expanded prereq is added, which just adds the shell override for the specific topic.  Since this is secondarily expanded, the shell is not overridden until the target is actually be executed. This along with the new wildcard target which resets the SHELL to the default, makes it so that prereq targets do not inherit this shell override, which is the default for variables of targets with respect to that target's prereqs.

There are some cases where $(shell) calls are resolved during the generation + execution of recipe, in which case those will end up using the overridden SHELL, the bash script.  In this case, we use an export variable, which only applies for the target given the enable_logging prereq to determine if we should log the pre and post log message.

The output of these targets should be same as before, as an example:

```
------------------- 2023-10-12T18:55:10.034+0000 Starting target=_output/bin/kube-vip/linux-amd64/kube-vip -------------------
(/home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/projects/kube-vip/kube-vip) $ /home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/build/lib/simple_create_binaries.sh /home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/projects/kube-vip/kube-vip /home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/projects/kube-vip/kube-vip/_output/bin/kube-vip/linux-amd64/kube-vip kube-vip 1.19 linux/amd64 "." "build" "" "-s -w -buildid= -extldflags -static " 0 "" "." "kube-vip"
Adding /go/go1.19/bin to PATH
Building binary(s): kube-vip
GOOS: linux
GOARCH: amd64
CGO_ENABLED: 0
GOCACHE: /root/.cache/go-build/1.19
(/home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/projects/kube-vip/kube-vip/kube-vip) $ go build -trimpath -a -ldflags -s -w -buildid= -extldflags -static  -o /home/prow/go/src/github.com/aws/eks-anywhere-build-tooling/projects/kube-vip/kube-vip/_output/bin/kube-vip/linux-amd64/kube-vip .
------------------- 2023-10-12T18:56:17.404+0000 Finished target=_output/bin/kube-vip/linux-amd64/kube-vip duration=67.373 seconds -------------------
```


build/lib/common.sh
```diff
@@ -149,7 +152,7 @@ function build::gather_licenses() {
   # data about each dependency to generate the amazon approved attribution.txt files
   # go-deps is needed for module versions
   # go-licenses are all the dependencies found from the module(s) that were passed in via patterns
-  build::common::echo_and_run go list -deps=true -json ./... | jq -s ''  > "${outputdir}/attribution/go-deps.json"
+  build::common::echo_and_run go list -deps=true -json ./... | jq -s '.'  > "${outputdir}/attribution/go-deps.json"
```
jaxesn, line 155: newer versions of jq do not like hte empty string. the `.` should be the same.

abhay-krishna (approved):
/lgtm
/approve

PR #2563, opened 2023-10-16
fixes attribution generation for cgo projects

jaxesn:
*Issue #, if available:*

*Description of changes:*

We had an issue with attribution generation where if it was generated in codebuild during checksum job it wouldnt include deps specific to cgo, but in prow in the nightly it would (im not 100% sure the diff there): https://github.com/aws/eks-anywhere-build-tooling/pull/2560#pullrequestreview-1680634433

This fixes that by export CGO_ENABLED when gathering deps for those projects.


abhay-krishna (approved):
/lgtm
/approve

PR #2567, opened 2023-10-17
Revert changes to CloudStack output glob

abhay-krishna:
The "fix" in #2565 actually failed Cloudstack builds because these builds, unlike OVA and Raw image builds, output their final build artifacts in the location `output/build-name/build-name`. OVA and Raw also generate similar files in a nested `build-name` directory but due to post processing (compression/archival), the final artifacts end up as `output/build-name.ova` and `output/build-name.gz`. No such post-processing occurs for cloudstack, so the output artifact is the QCOW2 image in `output/build-name/build-name` (without extension).

### DIFF
--- a/projects/aws/image-builder/builder/builder.go
+++ b/projects/aws/image-builder/builder/builder.go
@@ -87,6 +87,7 @@ func (b *BuildOptions) BuildImage() {
 
 		commandEnvVars = append(commandEnvVars, fmt.Sprintf("%s=%s", packerAdditionalFilesConfigFileEnvVar, additionalFilesConfigFile))
 	}
+	var outputImageGlobPattern string
 	if b.Hypervisor == VSphere {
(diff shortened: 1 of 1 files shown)


projects/aws/image-builder/builder/builder.go
```diff
@@ -233,10 +228,9 @@ func (b *BuildOptions) BuildImage() {
 		}
 
 		var buildCommand string
-		var outputImageGlobPattern string
 		switch b.Os {
 		case RedHat:
-			outputImageGlobPattern = "output/rhel-*"
+			outputImageGlobPattern = "output/rhel-*/rhel-*"
```
abhay-krishna, line 233: This is the fix, rest is just a minor refactor.

pokearu (approved):
/lgtm

PR #2580, opened 2023-10-18
minor image builder improvements

jaxesn:
*Issue #, if available:*

*Description of changes:*

- stop showing s3 upload progress. for image uploading it overloads the logs and its unnecessary
- extract the br raw image requires lz4
- avoid removing the downloaded files because they are intermediate targets. the fix is a bit ugly and this is a pretty minor case, it really only comes up when running locally multiple times, but feels like a low risk to avoid some "strange" behavior that I have to re-remember every time.
- improve the host interface for the packer host IP detection


abhay-krishna (approved):
/approve

PR #2585, opened 2023-10-18
adds support for buildx if buildctl is missing

jaxesn:
*Issue #, if available:*

*Description of changes:*

Recently I added a prereq for buildctl when building images. I had, incorrectly, assumed everyone was building images via the host since when using `make images` and pushing to some remote registry auth is most likely involved. @jacobweinstock informed me that he actually used to build images via the builder-base, the `make run-in-docker` targets, to avoid setting up buildctl/buildkit (he obviously did not read my building-locally doc....).  This new prereq broke his workflow since it no longer allows to build images in the builder base since the prereq is checking for the host to be setup and fails otherwise.

As a short history lesson, we have always used buildctl to build out images since from the early days of eks-d we were building images in prow, using buildkitd-rootless as a sidecar to the prow jobs.  Users do not usually have buildctl setup like they typically have docker.  Docker also nowadays both includes the `buildx` plugin and uses the BUILDKITD backend by default.  The reason this is nice is that buildx's cli arguments closely match those of buildctl, since its just frontend for buildkitd.

I actually added a (really basic) conversion script from buildctl to buildx for the eks-distro-build-tooling repo in this [PR](https://github.com/aws/eks-distro-build-tooling/pull/546).  The nice thing about buildx vs buildctl is you can configure it to point to multiple buildkitd instances. In the case of the eks-d builder-base, we actually point to an amd64 and an arm64 builder and buildx is smart enough to build the image on the proper instance which matches the arch.  In the case of eks-d builder base and minimal images this ends up having a huge effect on build times since building arm images on amd takes a significant hit going thru qemu.

We have not needed this behavior in this repo since 95% of our image builds are simple COPYs, very few are doing any RUN commands.  That said, we could utilize this conversion script to better support local building via `docker buildx` avoiding the need to setup buildctl.

### DIFF
--- a/build/lib/buildkit.sh
+++ b/build/lib/buildkit.sh
@@ -19,7 +19,60 @@ set -o pipefail
 
 BUILD_LIB_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/" && pwd -P)"
 
-CMD="buildctl"
+if [ "${USE_BUILDX:-}" == "true" ]; then
+    printf "\nBuilding with docker buildx\n" >&2
+
+    CMD="docker buildx"
+    ARGS=""
+    while test $# -gt 0; do
+        case "$1" in
+            --frontend)
+                shift
+                shift
+                ;;
+            --opt)
+                shift
+                if [[ $1 = filename* ]]; then
+                    ARGS+="--${1/filename/file} "
+                elif [[ $1 = "no-cache"* ]]; then
+                    ARGS+="--${1/no-cache/no-cache-filter} "
+                else
+                    ARGS+="--${1/:/ } "
+                fi
+                shift
+                ;;
+            --local)
+                shift
+                if [[ $1 = context* ]]; then
+                    ARGS+="${1/context=/} "
+                elif [[ $1 = dockerfile* ]]; then
+                    ARGS+="${1/dockerfile=/-f }/Dockerfile "
+                fi
+                shift
+                ;;
+            --export-cache)
+                shift
+                ARGS+="--cache-to $1 "
+                shift
+                ;;
+            --import-cache)
+                shift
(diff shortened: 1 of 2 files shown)


build/lib/buildkit.sh
```diff
@@ -19,7 +19,60 @@ set -o pipefail
 
 BUILD_LIB_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/" && pwd -P)"
 
-CMD="buildctl"
+if [ "${USE_BUILDX:-}" == "true" ]; then
+    printf "\nBuilding with with docker buildx\n" >&2
```
abhay-krishna, line 23: nit
```suggestion
    printf "\nBuilding with docker buildx\n" >&2
```

build/lib/buildkit.sh
```diff
@@ -34,25 +87,32 @@ fi
...
     log_file=$(mktemp)
     trap "rm -f $log_file" EXIT
-    if ! $CMD "$@" 2>&1 | tee $log_file; then        
+    if ! $CMD $ARGS 2>&1 | tee $log_file; then      
         if grep -q "blobs/uploads/\": EOF" $log_file ; then
             echo "******************************************************"
             echo "Ensure container registry and repository exists!!"
             echo "Try running make create-ecr-repos to create ecr repositories in your aws account."
             echo "******************************************************"
+        elif [ "$CMD" = "docker buildx" ] && grep -q "multiple platforms feature is currently not supported" $log_file; then
```
jaxesn, line 109: this is a bit annoying but the default docker buildx builder does not support multiarch builds. If people are building just single arches this isnt an issue, but added this message check to guide them on how to create a driver that will support multiple.  i tested this on an al2 instance.

PR #2605, opened 2023-10-24
Fix unschedulable Tinkerbell stack when no worker nodes

chrisdoherty4:
Closes https://github.com/aws/eks-anywhere-internal/issues/1953.

If all worker nodes elgible to run Tinkerbell stack workloads go offline the cluster and Workload clusters cannot recover failed nodes. This change allows the Tinkerbell stack to run on control plane nodes but prefer Worker nodes at scheduling time.

The patch was tested against a 1 control plane 1 worker node cluster. The Helm chart was installed using existing values, then the worker node was drained. I observed the Tinkerbell stack move to the control plane node. Similarly, when a pod is deleted it preferences the worker node (when untainted).


jacobweinstock (approved):

chrisdoherty4:
Found a problem with VIP advertisment. Will need to make a small adjustment.

/hold

PR #2615, opened 2023-10-28
support pushing/pulling from private s3

jaxesn:
*Issue #, if available:*

*Description of changes:*

We do not need to use this for anything right during our build process, but we may eventually want to enable building in CI without marking all build artifacts public.

I was using this to test building the entire stack in my aws account pointed at a private s3 bucket. I was able to build the entire stack this way without needing to pull anything from the build account bucket.


abhay-krishna (approved):
/lgtm
/approve

PR #2628, opened 2023-11-02
better handle no golang version when cleaning up go cache directories

jaxesn:
*Issue #, if available:*

*Description of changes:*

The nightly attribution generation has been failing since my docker changes.  I had changed the use_go_version function to return early if the supplied go_version was empty. Before it would have just defaulted to the default go version in the container.  Since it never gets set, the GOCACHE folder was being set to `/root/.cache/go-build` instead of `/root/.cache/go-build/{GO_VERSION}`.  This not a big deal since that project isnt doing any building anyway.  

The problem comes when in prow we mount` /root/.cache/go-build` so when clean-go-cache tried to delete it, it cant because its not really a folder. This changes that code to check for a go_version before trying to delete those go folders.

Since i was in here, i also tweaked the way the trace shell works so it can be used when targets override to use logging or docker.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -13,9 +13,9 @@ BUILD_LIB=$(BASE_DIRECTORY)/build/lib
 OUTPUT_BIN_DIR?=$(OUTPUT_DIR)/bin/$(REPO)
 
 SHELL_TRACE?=false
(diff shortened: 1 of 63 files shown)


abhay-krishna (approved):
/approve

PR #2636, opened 2023-11-06
gives a better error message when downloading dependencies fails

jaxesn:
*Issue #, if available:*

*Description of changes:*

When downloading deps fails its pretty silent. This adds a better error message to guide the dev.  Also improved the rerunable of the targets by delaying when the extract folder is created so that if download fails and you rerun, it will try again versus thinking it was already done due to the folder existing.


abhay-krishna (approved):
/approve

PR #2637, opened 2023-11-07
downloads iso before calling qemu

jaxesn:
*Issue #, if available:*

*Description of changes:*

In CI we ocassionally will see a timeout in qemu when it tries to download the iso, for either redhat or ubuntu. This changes the flow a bit to download the iso first before calling qemu.


vignesh-goutham (approved):
/approve

PR #2642, opened 2023-11-07
supports s3 path prefixes in bucket variable

jaxesn:
*Issue #, if available:*

*Description of changes:*

When building privately or in the case of staging bundle build, wed like to be able to push to and pull from a clean folder on s3. This will allow us to ensure that when we build the entire stack we are building all the deps as needed and not potentially pulling older builds.

The way this is setup is to support supplying a path/prefix in ARTIFACTS_BUCKET. for ex

`export ARTIFACTS_BUCKET=s3://my-bucket/staging/release/1`

Another option would be to introduce a new var, ARTIFACTS_BUCKET_PREFIX.  This approach may be slightly more clear, but there are a number of places wed have to handle this in the code today.  If we really wanted to have a second var we could something like:

```
export ARTIFACTS_BUCKET=s3://my-bucket
export ARTIFACTS_BUCKET_PREFIX=/staging/build/1
```

Then in Common.mk do:

`ARTIFACTS_BUCKET+=$(ARTIFACTS_BUCKET_PREFIX)`

This approach has the benefit of 2 vars but to all the existing code, ARTIFACTS_BUCKET has the full path that it needs.

The utilmate plan here is to change the bundle pipeline to generate a new, unique, s3 prefix/folder so that all the builds in that pipeline push and pull from that folder, instead of the `latest` folder.

### DIFF
--- a/build/lib/common.sh
+++ b/build/lib/common.sh
@@ -99,7 +99,7 @@ function build::common::generate_shasum() {
 
 function build::common::upload_artifacts() {
   local -r artifactspath=$1
-  local -r artifactsbucket=$2
+  local -r artifactsbucket="${2%/}" # remove trailing slash if it exists
   local -r projectpath=$3
   local -r buildidentifier=$4
   local -r githash=$5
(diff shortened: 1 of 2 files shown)


abhay-krishna (approved):
/approve

PR #2645, opened 2023-11-09
adds check for qemu and updates docs a bit

jaxesn:
*Issue #, if available:*

*Description of changes:*

In the case of cross-compiling, we use the builder base but the platform may not always match the host.  For this to work you need qemu installed, which we do all over and have in the docs.  This adds a check for this and a message to guide the user.

Interestingly, I was going to do the same for buildctl and buildx, but buildkit (which powers both) seems to ship with its own qemu and so you do not actually have to run the binfmt install... this is news to me....

I also updated the building locally doc a bit, reorg mainly but a few update items.

### DIFF
--- a/build/lib/common.sh
+++ b/build/lib/common.sh
@@ -493,6 +493,24 @@ function build::common::copy_if_source_destination_different() {
   cp -rf $source $destination
 }
 
+function build::common::check_for_qemu() {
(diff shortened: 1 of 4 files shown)


abhay-krishna (approved):
/approve

PR #2649, opened 2023-11-10
Image-builder download artifacts command

vignesh-goutham:
*Description of changes:*
Adds a new download artifacts command that will download the required artifacts needed for a combination of EKS-A release and EKS-D release channel. These artifacts will be required to host on private internal servers for air-gapped image building.

Also upgrades go from 1.18 -> 1.21


projects/aws/image-builder/cmd/downloadartifacts.go
```diff
@@ -0,0 +1,32 @@
...
+	"log"
+)
+
+var downloadArtifactsCmd = &cobra.Command{
+	Use:   "artifacts",
+	Short: "Downloads artifacts for airgapped builds",
+	Long:  "Downloads EKS-D and EKS-A artifacts used to build EKS-A Node images in airgapped environment",
+	Run: func(cmd *cobra.Command, args []string) {
+		// Validate Inputs
+		if bo.EKSAReleaseVersion == "" {
```
abhay-krishna, line 14: Can we use this method in `init()` to validate the flag is provided?https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagRequired

vignesh-goutham: good catch. I dont know what I was thinking.

projects/aws/image-builder/builder/artifacts.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	var fullKubeVersion, eksDBaseUrl string
+	for _, component := range releases.Status.Components {
+		if component.Name == "kubernetes" {
+			fullKubeVersion = component.GitTag
+			log.Printf("Full Kube Version: %s\n", fullKubeVersion)
+
+			for _, asset := range component.Assets {
+				if asset.Name == "bin/linux/amd64/kube-apiserver.tar" {
```
jaxesn, line 101: i know we do not support building arm images, but can we go ahead and parametrize `amd64`, even if we just hardcode it vs letting the user supply?  One less we have to think about later

jaxesn (changes requested):
this seems pretty reasonable. a test or two would probably good..

jaxesn (approved):

PR #2658, opened 2023-11-14
run emmissary as a batch build in CI like harbor

jaxesn:
*Issue #, if available:*

*Description of changes:*

We are hitting timeouts in CI due to these images taking a while to build. We broke harbor out for the same reason. I followed the same pattern here.

- [ ] Change CDK to support batch builds for emissary


abhay-krishna (approved):
/approve

PR #2670, opened 2023-11-17
Bump harbor version

d8660091:
## upstream changelogs
harbor: https://github.com/goharbor/harbor/compare/v2.7.1...v2.9.1
harbor-helm: https://github.com/goharbor/harbor-helm/compare/v1.11.1...v1.13.1

## upstream golang source of truth
https://github.com/goharbor/harbor/blob/main/src/go.mod#L3

## explanations of changed/removed patches
0001-add-digest-namespace-and-imagepullsecret-support.patch: changed based on the upstream change (chartmuseum, notary removed)
0002-generate-swagger-models-and-restapis.patch: patch is regenerated from "make gen_apis"
0003-security-patch.patch: no longer needed, upstream has patched dependencies


projects/goharbor/harbor/docker/linux/harbor-portal/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG BASE_IMAGE # https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nginx
 
-FROM node:16.10.0 as nodeportal
+FROM node:16.18.0 as nodeportal
```
abhay-krishna, line 3: Can we use `https://gallery.ecr.aws/docker/library/node` instead of the node image from DockerHub?

d8660091: sure

PR #2671, opened 2023-11-17
use tee to fix digest to tmpfile

jaxesn:
*Issue #, if available:*

*Description of changes:*

I *think* I had originally changed this because for some reason tee itself does not work for me on my mac, or my linux instances. It appears that without using tee it does not work in codebuild.. not sure if its a difference in the helm version?


abhay-krishna (approved):
/lgtm
/approve

PR #2679, opened 2023-11-20
Bump capi v1.5.3 and cert-manager v1.13.1

g-gaston:
*Description of changes:*
Notable changes:
The first patch 0001 (for BR) changes the go.mod. Modifying the go.mod is annoying bc it increased the changes of having conflicts on updates (like we did with this patch).

It was just adding a dependency on a yaml module. But that was unnecessary, we can use the k8s yaml package that it's already imported by capi. I removed the go.mod change, changed the import statement and added a test to validate the change (that I ran before and after the code change).

This should make future updates easier.


abhay-krishna (approved):
/approve

PR #2681, opened 2023-11-21
Support release-branched projects and passing project name in README version check

abhay-krishna:
* Allow passing project name to readme update script to update version for single project
* Fix bug in echo command that caused the `generate` Make target to be run during in case of error.
* Block merge when project README is not up to date.

### DIFF
--- a/build/lib/readme_check.sh
+++ b/build/lib/readme_check.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#set -x
+# set -x
 set -o errexit
 set -o nounset
 set -o pipefail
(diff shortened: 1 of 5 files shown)


build/lib/readme_check.sh
```diff
@@ -23,11 +23,17 @@ source "${SCRIPT_ROOT}/common.sh"
 
 RETURN=0
 SED=$(build::find::gnu_variant_on_mac sed)
+PROJECT=${1:-}
 
-for GIT_TAG_FILE in projects/*/*/GIT_TAG
+GIT_TAG_FILES="projects/*/*/GIT_TAG"
```
jaxesn, line 28: does this work for projects whcih are released branched?

abhay-krishna: I think even earlier, we weren't checking for release branched projects since for those, the README format is slightly different so for now, manual updates are required.

abhay-krishna: Added support for release branched projects.

jaxesn (approved):

PR #2682, opened 2023-11-21
standardize batch builds for archd images

jaxesn:
*Issue #, if available:*

*Description of changes:*

We had two projects, harbor and emissary, which we had broken out into a batch build by image platform due to how long they took to build on a single host.  The reason they took so long is when building images that do more than just COPY can take quiet away when running for the arch that does not match the host, due to the virtualization through qemu.

So far I had just hand made the batch buildspecs for these two projects.  We just added the upgrader project which started to segfault in CI due to qemu when building the amd image on our arm host.  I *think* we could downgrade qemu to 6.x from 7.x, 6.x appears to be more stable, however probably the safest thing to do is run these image builds on a matching host arch.  

We ran into this same thing in eks-d where we build the minimal images in prow and ended up doing the same thing.  

This PR adds the automation needed to make generating the build specs by image platform easier.


abhay-krishna (approved):
/approve

PR #2688, opened 2023-11-27
refactor ansible retry and timeout after 1h

jaxesn:
*Issue #, if available:*

*Description of changes:*

We had a failure in codebuild which may be the retry code, but I actually think its ansible just stalling. This refactors this code a bit and also adds a timeout to ansible. 1h was picked because the individual provisioners have a 30 min time out.


abhay-krishna (approved):
/approve

PR #2736, opened 2023-12-06
Skip leftover packages in upgrade job, handle up-to-date remote when pushing

abhay-krishna:
* Remove some packages projects that were left out in a previous PR (skipping `aws/eks-anywhere-packages` because it is maintained and released by us and we will bump it ourselves immediately).
* Handle Git push errors when pushing to an up-to-date remote.
* Apply `do-not-merge/hold` labels through Prow command instead of directly through GitHub API since it requires elevated permissions which the bot doesn't have.


pokearu (approved):
/lgtm

projects/aws/eks-anywhere-packages/Makefile
```diff
@@ -40,6 +40,8 @@ HELM_CHART_NAMES=eks-anywhere-packages-migrations eks-anywhere-packages-crds eks
 
 BUILDSPEC_COMPUTE_TYPE=BUILD_GENERAL1_LARGE
 
+EXCLUDE_FROM_UPGRADE_BUILDSPEC=true
```
jaxesn, line 43: Id probably keep this along with all the packages since what would it hurt? But we can review all these skipped ones as we go and use this automation for others

abhay-krishna: Yeah the idea is to revisit these once the job stabilizes.

PR #2770, opened 2023-12-14
Image-builder: Update kernel and keep only 1 kernel on redhat machines

vignesh-goutham:
*Description of changes:*
Image builder was excluding kernel updates till now. With this change, image builder will update kernels and reboot into the new kernel. Post reboot, if the OS has more than 1 kernel, this change will remove all non-live kernels running on the RHEL machine.

*Testing*
Running image builder with changes output
```
    qemu: TASK [node : Get the list of installed kernels] ********************************
    qemu: changed: [default]
    qemu:
    qemu: TASK [node : Get the current running kernel] ***********************************
    qemu: changed: [default]
    qemu:
    qemu: TASK [node : Set the number of kernels to keep] ********************************
    qemu: ok: [default]
    qemu:
    qemu: TASK [node : Filter out the current kernels] ***********************************
    qemu: ok: [default]
    qemu:
    qemu: TASK [node : Remove old kernels] ***********************************************
    qemu: ok: [default] => (item=['4.18.0-372.9.1.el8'])
```

/hold 

### DIFF
--- a/projects/kubernetes-sigs/image-builder/patches/0028-Remove-old-kernels.patch
+++ b/projects/kubernetes-sigs/image-builder/patches/0028-Remove-old-kernels.patch
@@ -1,0 +1,82 @@
+From 39a06391908f873f747f22d49682ec7d6009a7af Mon Sep 17 00:00:00 2001
+From: Vignesh Goutham Ganesh <vgg@amazon.com>
+Date: Thu, 14 Dec 2023 15:36:28 -0600
+Subject: [PATCH] Remove old kernels
+
+Signed-off-by: Vignesh Goutham Ganesh <vgg@amazon.com>
+---
+ images/capi/ansible/roles/node/tasks/main.yml |  3 ++
+ .../capi/ansible/roles/node/tasks/redhat.yml  | 32 +++++++++++++++++++
+ .../capi/ansible/roles/setup/tasks/redhat.yml |  2 +-
+ 3 files changed, 36 insertions(+), 1 deletion(-)
+ create mode 100644 images/capi/ansible/roles/node/tasks/redhat.yml
+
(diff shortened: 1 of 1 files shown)


projects/kubernetes-sigs/image-builder/patches/0028-Remove-old-kernels.patch
```diff
@@ -0,0 +1,90 @@
...
++  ansible.builtin.command: "uname -r"
++  register: current_kernel
++
++- name: Set the number of kernels to keep
++  set_fact:
++    kernels_to_keep: 1
++
++- name: Filter out the current kernels
++  set_fact:
++    old_kernels: "{{ installed_kernels.stdout_lines | difference([current_kernel.stdout]) | reverse | slice(kernels_to_keep)|list }}"
```
abhay-krishna, line 67: Based on my understanding of this command, this is what's happening
For example:
`installed_kernels` = `[4.18, 5.10, 5.15]`
`current_kernel` = `5.15`
`kernels_to_keep` = `1`
`installed_kernels.stdout_lines | difference([current_kernel.stdout])` = `[4.18, 5.10]`
`... | reverse` = `[5.10, 4.18]`
`... | slice(kernels_to_keep) | list` = `[5.10]` ===> `old_kernels` = `[5.10]`

if my understanding is correct, aren't we skipping removing the 4.18 kernel in the above case, or are we just making the assumption that exactly one old kernel will be there besides the current one? Or am I interpreting it incorrectly?

vignesh-goutham: Yes, there will always be only 1 extra kernel present on the image. But I missed adding my changes, this command is not needed to be this extensive. My test OS had 3 kernels for some reason which is not the case with base ISOs. Updating PR in a bit.

abhay-krishna (approved):
/approve

PR #2799, opened 2024-01-04
Detect PACKER_ACTIVE_INTERFACE with "/proc/net/route" if possible

guessi:
*Issue #, if available:*

*Description of changes:*

While running `setup_packer_configs.sh` with Ubuntu OS, user might have `route` command not found issue while checking with `$(route |awk '/^default/{print $NF}')`. Consider not all OS variant comes with `route` command, it could be replaced with the `ip` instead, which is used in elsewhere in the same script.

Furthermore, PR #2219 was initially aim to support Ubuntu 22.04 ( see #2219 ), but it also unblock all the Ubuntu versions. If the script is aim to support all Ubuntu versions, it would better to have `ens5` supported.

image

### DIFF
--- a/projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
+++ b/projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
@@ -160,20 +160,53 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
         if [ -z $PACKER_ACTIVE_INTERFACE ]; then
             echo "Finding interface for Packer temporary HTTP server"
             if [ "$(uname -s)" = "Linux" ]; then
-                if command -v route &> /dev/null; then
-                    PACKER_ACTIVE_INTERFACE=$(route |awk '/^default/{print $NF}')
+                # Linux distros nowadays should have /proc/net/route with route info as follow,
+                #
+                # (Amazon Linux 2, RHEL, Ubuntu, ...)
+                # $ cat /proc/net/route
+                # Iface    Destination    Gateway     Flags       RefCnt    Use    Metric    Mask        MTU    Window    IRTT
+                # eth0     00000000       01201FAC    0003        0         0      0         00000000    0      0         0
+                # ...
+                #
+                # With systemd v197 (or newer) comes with predictable interface name feature so the interface name,
+                # a reliable mechanism to determine which interface have default route associated would be required.
+                #
+                # However, not all Linux distro comes with binary tools like `ip` or `route` (e.g., Ubuntu don't have `route` command by default)
+                # so here we leverage the info from `/proc/net/route` which is more reliable and trustworthy.
+                #
+                # References:
+                # - https://systemd.io/PREDICTABLE_INTERFACE_NAMES/
+                # - https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
+                # - https://en.wikipedia.org/wiki/Consistent_Network_Device_Naming
+                IFACE_WITH_DEFAULT_ROUTE=$(cat /proc/net/route | awk '$2 == "00000000" && $8 == "00000000" { print $1 }' | head -1) # Only the one with lowest "Metric" route would be taken.
+                if [ -n $IFACE_WITH_DEFAULT_ROUTE ]; then
+                    PACKER_ACTIVE_INTERFACE=$IFACE_WITH_DEFAULT_ROUTE
                 else
-                    INTERFACES=($(ls /sys/class/net))
-                    for interface in "${INTERFACES[@]}"; do
-                        if [ "$interface" = "eth0" ] || [ "$interface" = "en0" ] || [ "$interface" = "eno1" ]; then
-                            PACKER_ACTIVE_INTERFACE=$interface
-                            echo "Found interface: $interface"
+                    for interface in $(cat /proc/net/route | awk '! /Iface/{print $1}' | sort | uniq); do
+                        predictable_ifnames=(
+                            "eth0"
+                            "en0"
+                            "eno1"
+                            "ens5"
+                        )
+
+                        matched=0
+                        for iface in "${predictable_ifnames[@]}"; do
+                            if [[ "$iface" == "$interface" ]]; then
+                                matched=1
+                                PACKER_ACTIVE_INTERFACE=$iface
+                                echo "Found interface: $iface"
+                                break
+                            fi
+                        done
+
+                        if [[ $matched -eq 1 ]]; then
                             break
                         fi
                     done
                 fi
             elif [ "$(uname -s)" = "Darwin" ]; then
-                PACKER_ACTIVE_INTERFACE=$(route -n get default | grep 'interface:' | grep -o '[^ ]*$')
+                PACKER_ACTIVE_INTERFACE=$(route -n get default | awk '/interface: /{print $2}')
             fi
         fi
 
@@ -183,7 +216,7 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
         fi
 
         if [ "$(uname -s)" = "Linux" ]; then
-            PACKER_HTTP_SERVER_IP=$(ip a l $PACKER_ACTIVE_INTERFACE | awk '/inet / {print $2}' | cut -d/ -f1)
+            PACKER_HTTP_SERVER_IP=$(ip addr show dev $PACKER_ACTIVE_INTERFACE | awk '/inet / {print $2}' | cut -d/ -f1)
         elif
             [ "$(uname -s)" = "Darwin" ]; then
             PACKER_HTTP_SERVER_IP=$(ifconfig $PACKER_ACTIVE_INTERFACE | awk '/inet / {print $2}' | cut -d/ -f1)


projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,12 +160,16 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                if command -v ip &> /dev/null; then
+                    PACKER_ACTIVE_INTERFACE=$(ip route show default | awk '/default/ {print $5}')
                 else
                     INTERFACES=($(ls /sys/class/net))
                     for interface in "${INTERFACES[@]}"; do
-                        if [ "$interface" = "eth0" ] || [ "$interface" = "en0" ] || [ "$interface" = "eno1" ]; then
+                        # Systemd v197 (or newer) comes with predictable interface name and under Ubuntu system on the Cloud, it would be "ens5".
+                        # - https://systemd.io/PREDICTABLE_INTERFACE_NAMES/
+                        # - https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
+                        if [ "$interface" = "eth0" ] || [ "$interface" = "en0" ] || [ "$interface" = "eno1" ] || [ "$interface" = "ens5" ]; then
```
abhay-krishna, line 172: I'm okay to keep adding to this list, but I think for readability's sake, it would be better to define an array and search from it.
```bash
predictable_interface_names=(
    eth0
    en0
    eno1
    ens5
)

if [[ ${predictable_interface_names[@]} =~ $interface ]]; then
...
```

guessi: Sure thing, it's much easier to read ( or leaving comments for why we list such item )

guessi: #### Script

```
$ cat test.sh
#!/bin/bash

interface=$1

predictable_interface_names=(
        "eth0"
        "en0"
        "eno1"
        "ens5"
)

if [[ ${predictable_interface_names[@]} =~ $interface ]]; then
    echo "found in array"
else
    echo "not in array"
fi
```

#### Test 1: ✅ 
```
$ bash -x test.sh eth0
+ interface=eth0
+ predictable_interface_names=("eth0" "en0" "eno1" "ens5")
+ [[ eth0 en0 eno1 ens5 =~ eth0 ]]
+ echo 'found in array'
found in array
```

#### Test 2: ✅ 
```
$ bash -x test.sh eth1
+ interface=eth1
+ predictable_interface_names=("eth0" "en0" "eno1" "ens5")
+ [[ eth0 en0 eno1 ens5 =~ eth1 ]]
+ echo 'not in array'
not in array
```

#### Test 3: ❌ 
```
% bash -x test.sh eth
+ interface=eth
+ predictable_interface_names=("eth0" "en0" "eno1" "ens5")
+ [[ eth0 en0 eno1 ens5 =~ eth ]]
+ echo 'found in array'
found in array
```

guessi: Just a second, I can handle it, let me work on it...

guessi: Updated.

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,14 +160,33 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                                PACKER_ACTIVE_INTERFACE=$iface
+                                echo "Found interface: $iface"
+                                break
+                            fi
+                        done
+
+                        if [[ $matched -eq 1 ]]; then
                             break
                         fi
                     done
```
guessi, line 192: Test script

```
% cat test.sh
#!/bin/bash

#
# Expected to have first matched "ens5"
#

INTERFACES=("eth" "etx1" "ens5" "eth0")

for interface in "${INTERFACES[@]}"; do
   predictable_interface_names=(
       "eth0"
       "en0"
       "eno1"
       "ens5"
   )

   matched=0
   for iface in "${predictable_interface_names[@]}"; do
       if echo $iface | grep -o -q -E ^$interface$; then
           matched=1
           PACKER_ACTIVE_INTERFACE=$iface
           echo "Found interface: $iface"
           break
       fi
   done

   if [[ $matched -eq 1 ]]; then
        break
    fi
done

echo "PACKER_ACTIVE_INTERFACE: ${PACKER_ACTIVE_INTERFACE}"
```

Test result:
```
$ bash ./test.sh
Found interface: ens5
PACKER_ACTIVE_INTERFACE: ens5
```

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,14 +160,33 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                        predictable_interface_names=(
+                            "eth0"
+                            "en0"
+                            "eno1"
+                            "ens5"
+                        )
+
+                        matched=0
+                        for iface in "${predictable_interface_names[@]}"; do
+                            if echo $iface | grep -o -q -E ^$interface$; then
```
abhay-krishna, line 181: technically we could just do
```
if [[ "$iface" == "$interface" ]];
```

guessi: Oops... yes 🤣

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,14 +160,33 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
         if [ -z $PACKER_ACTIVE_INTERFACE ]; then
             echo "Finding interface for Packer temporary HTTP server"
             if [ "$(uname -s)" = "Linux" ]; then
-                if command -v route &> /dev/null; then
-                    PACKER_ACTIVE_INTERFACE=$(route |awk '/^default/{print $NF}')
+                # NOTE: not all Ubuntu variant comes "route" command by default, better to detect with "ip".
+                if command -v ip &> /dev/null; then
```
jaxesn, line 164: any reason not to just do both, if route exists use it otherwise check for ip?

jaxesn (commented):
I love this submission!  Thanks so much for the detailed description.

This script has never been a particular favorite of mine... When I changed this in my 2204 PR, it really didnt have anything to do with 2204 support, since that was more about adding support to build node images based on 2204 vs this process runs on some kind of "admin machine".

I should have broken that change out to be more clear that it was unrelated.  That said, I *think* the actual PR where I introduced the route call was #2580 which does not give a lot of detail as to why, my bad :(.

I love the changes to the known interfaces, thanks @abhay-krishna for those reviews.  For the route -> ip change, is there any value in just doing both?  Like I said, I dont recall what was going thru my mind when i added the route call so I can defend it as being necessary, just wanting to get your thoughts?

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
abhay-krishna: +1

guessi: It's simply because we had `ip` elsewhere in the script. I can do it for both.... should I ?

Or maybe we should rely on info from /sys/class/... so that we could get rid of binary deps, WDYT?

guessi: Let me check if there is a way to get default route without binary requirements. I will get you back later.

guessi: How about this?

### Checking with `ip`

```
$ ip route show default
default via 192.168.96.1 dev eth0
```

### Checking with `route`
```
$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.96.1    0.0.0.0         UG    0      0        0 eth0
169.254.169.254 0.0.0.0         255.255.255.255 UH    0      0        0 eth0
169.254.170.23  0.0.0.0         255.255.255.255 UH    0      0        0 pod-id-link0
192.168.96.0    0.0.0.0         255.255.224.0   U     0      0        0 eth0
192.168.119.120 0.0.0.0         255.255.255.255 UH    0      0        0 eni7e18413c9aa
192.168.122.116 0.0.0.0         255.255.255.255 UH    0      0        0 enia4662b17240
```

### Checking with `/proc/net/route`
```
$ cat /proc/net/route
Iface	Destination	Gateway 	Flags	RefCnt	Use	Metric	Mask		MTU	Window	IRTT
eth0	00000000	0160A8C0	0003	0	0	0	00000000	0	0	0
eth0	FEA9FEA9	00000000	0005	0	0	0	FFFFFFFF	0	0	0
pod-id-link0	17AAFEA9	00000000	0005	0	0	0	FFFFFFFF	0	0	0
eth0	0060A8C0	00000000	0001	0	0	0	00E0FFFF	0	0	0
eni7e18413c9aa	7877A8C0	00000000	0005	0	0	0	FFFFFFFF	0	0	0
enia4662b17240	747AA8C0	00000000	0005	0	0	0	FFFFFFFF	0	0	0
```

### Extracting default route info from `/proc/net/route`
```
$ cat /proc/net/route | awk '$2 == "00000000" { print $1 }'
eth0
```

Or even more stricter
```
$ cat /proc/net/route | awk '$2 == "00000000" { print $1 }' | head -1
eth0
```

guessi: Test with `Ubuntu 22.04`

```
ubuntu@ip-172-31-42-28:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
```

```
ubuntu@ip-172-31-42-28:~$ ip route show default
default via 172.31.32.1 dev ens5 proto dhcp src 172.31.42.28 metric 100
```

```
ubuntu@ip-172-31-42-28:~$ route -n
Command 'route' not found, but can be installed with:
sudo apt install net-tools
```

```
ubuntu@ip-172-31-42-28:~$  cat /proc/net/route
Iface	Destination	Gateway 	Flags	RefCnt	Use	Metric	Mask		MTU	Window	IRTT
ens5	00000000	01201FAC	0003	0	0	100	00000000	0	0	0
ens5	02001FAC	01201FAC	0007	0	0	100	FFFFFFFF	0	0	0
ens5	00201FAC	00000000	0001	0	0	100	00F0FFFF	0	0	0
ens5	01201FAC	00000000	0005	0	0	100	FFFFFFFF	0	0	0
```

```
ubuntu@ip-172-31-42-28:~$ cat /proc/net/route | awk '$2 == "00000000" { print $1 }'
ens5
```

```
ubuntu@ip-172-31-42-28:~$ cat /proc/net/route | awk '$2 == "00000000" && $8 == "00000000" { print $1 }' | head -1
ens5
```

guessi:
**_It might be a little bit off-topic, could be discussed in separated issue/PR._**

I think the Ubuntu version check could be done with regrex easier?

Here the test output with `Ubuntu 22.04.3 LTS`

```
ubuntu@ip-172-31-42-28:~$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

Extract the info we need
```
ubuntu@ip-172-31-42-28:~$ cat /etc/os-release | awk -F'"' '/PRETTY_NAME/{print$2}'
Ubuntu 22.04.3 LTS
```

As EKS-A supported versions are clearly to be "LTS", we could leverage regex to detect easier
```
ubuntu@ip-172-31-42-28:~$ cat /etc/os-release | awk -F'"' '/PRETTY_NAME/{print$2}' | grep -E 'Ubuntu 2[02]\.04.* LTS' # detect for Ubuntu 20.04 LTS or Ubuntu 22.04 LTS only
Ubuntu 22.04.3 LTS
```

Pitfalls need to take care of ...

* Ubuntu 22.04 LTS (initial release)
* Ubuntu 22.04.3 LTS (security update releases)

Does EKS-A need to take ESM into consideration?

* https://ubuntu.com/about/release-cycle
* https://ubuntu.com/security/esm

jaxesn:
I'm not sure I am following the Ubuntu version part? Where does, or would, this come into play?

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
jaxesn: This seems reasonable, thanks for testing on multiple platforms!  Technically we support building images on al2/ubuntu 20 + 22 and rhel, any chance you can try on rhel real quick? if not im sure its fine.

I would change the code slightly to do your new check and then if its empty for any reason fallback to the known interface names, instead of like we do now checking if `ip` is available and only falling back if its not.

guessi:
> I'm not sure I am following the Ubuntu version part? Where does, or would, this come into play?

It's about the code block I mentioned at the description of the PR initially.

jaxesn:
ah ok. that code is kind of a different case since those are the params for the type of node image we are building vs what the host OS is. So the 2004/ubuntu/rhel is all passed into that script, effectively hardcoded in our CI builds.

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
jaxesn: verified on an rhel node that your code works there too!  lets move forward with that approach.

guessi: Thanks for helping on verifying with RHEL, will submit an update for this later.

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -183,7 +216,7 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
         fi
 
         if [ "$(uname -s)" = "Linux" ]; then
-            PACKER_HTTP_SERVER_IP=$(ip a l $PACKER_ACTIVE_INTERFACE | awk '/inet / {print $2}' | cut -d/ -f1)
+            PACKER_HTTP_SERVER_IP=$(ip addr show dev $PACKER_ACTIVE_INTERFACE | awk '/inet / {print $2}' | cut -d/ -f1)
```
guessi, line 219: Try to improve readability with `ip route show dev $iface`

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,20 +160,53 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                        done
+
+                        if [[ $matched -eq 1 ]]; then
                             break
                         fi
                     done
                 fi
             elif [ "$(uname -s)" = "Darwin" ]; then
-                PACKER_ACTIVE_INTERFACE=$(route -n get default | grep 'interface:' | grep -o '[^ ]*$')
+                PACKER_ACTIVE_INTERFACE=$(route -n get default | awk '/interface: /{print $2}')
```
guessi, line 209: `awk` should be enough and easier to read.

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,20 +160,53 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                IFACE_WITH_DEFAULT_ROUTE=$(cat /proc/net/route | awk '$2 == "00000000" && $8 == "00000000" { print $1 }' | head -1) # Only the one with lowest "Metric" route would be taken.
+                if [ -z $IFACE_WITH_DEFAULT_ROUTE ]; then
+                    PACKER_ACTIVE_INTERFACE=$IFACE_WITH_DEFAULT_ROUTE
                 else
-                    INTERFACES=($(ls /sys/class/net))
-                    for interface in "${INTERFACES[@]}"; do
-                        if [ "$interface" = "eth0" ] || [ "$interface" = "en0" ] || [ "$interface" = "eno1" ]; then
-                            PACKER_ACTIVE_INTERFACE=$interface
-                            echo "Found interface: $interface"
+                    for interface in $(cat /proc/net/route | awk '! /Iface/{print $1}' | sort | uniq); do
```
guessi, line 185: Might have more than one entries for single interface, need to dedup here.

projects/kubernetes-sigs/image-builder/build/setup_packer_configs.sh
```diff
@@ -160,20 +160,53 @@ if [ "$IMAGE_FORMAT" = "ova" ] && \
...
+                #
+                # However, not all Linux distro comes with binary tools like `ip` or `route` (e.g., Ubuntu don't have `route` command by default)
+                # so here we leverage the info from `/proc/net/route` which is more reliable and trustworthy.
+                #
+                # References:
+                # - https://systemd.io/PREDICTABLE_INTERFACE_NAMES/
+                # - https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
+                # - https://en.wikipedia.org/wiki/Consistent_Network_Device_Naming
+                IFACE_WITH_DEFAULT_ROUTE=$(cat /proc/net/route | awk '$2 == "00000000" && $8 == "00000000" { print $1 }' | head -1) # Only the one with lowest "Metric" route would be taken.
+                if [ -z $IFACE_WITH_DEFAULT_ROUTE ]; then
```
abhay-krishna, line 182: Shouldn't this be `-n`?
```suggestion
                if [ -n $IFACE_WITH_DEFAULT_ROUTE ]; then
```

guessi: Sorry for the delay, this issue just fixed.

guessi:
Looks like all tests on Prow passed 🎉 

Let me know if there's anything need to be revised. Happy to continue on it if required.

jaxesn (approved):
Awesome work!

abhay-krishna:
> Looks like all tests on Prow passed 🎉
> 
> Let me know if there's anything need to be revised. Happy to continue on it if required.
> 
> _Nit:_ Originally the PR was aim to replace `route` with `ip` but it's not for now. The branch name for the PR still `replace-route-with-ip` 😆 ... if it required a new PR, it's totally, just let me know.

Thanks for the contribution @guessi! No issues with the branch name, happy to merge it as is.

/lgtm
/approve

PR #2806, opened 2024-01-08
fixes kindnetd manifest (again)

jaxesn:
*Issue #, if available:*

*Description of changes:*

This should be a better fix for the kindnetd manifest situation.

*Background* We handle the kind image and kindnetd a bit special compared to other images across the platform.  Originally we only supported cilium on any eks-a cluster, including docker/capd ones.  When adding (experimental) arm64 support our build of cilium only support amd64.  To work around I added basic support for kindnetd cni when creating a capd cluster.  

Kind images embed a number of container images, to support offline workflows as well as just less runtime downloading of containers.  EKS-A has always followed model so the kindnetd cni container image, along with its manifest are embedded in the kind node image. Again this the upstream behavior which we just inherit.  Its important to remember that we use the same kind image/process for our bootstrap cluster as we do a full capd based cluster.

We have always used kindnetd when creating a bootstrap cluster, I say always but really originally we used cilium pre-ga but ran into issues which looked like ebpf overwriting each other between the bootstrap cluster and the actual eks-a/capd cluster.  At that time we decided to use kindnetd on bootstrap and cilium on capd.  

Due to the way our releases and image tagging work, we do not know the image tag of an image at build time, its decided much later on during releases.  To allow kindnetd to work in the bootstrap cluster, we, at build time, modify the embedded kindnetd manifest file to point to an image tag, such as `public.ecr.aws/eks-anywhere/kubernetes-sigs/kind/kindnetd:v0.18.0` but you will notice this image does not exit in [ecr](https://gallery.ecr.aws/eks-anywhere/kubernetes-sigs/kind/node).  This is admittedly confusing and odd, but we rely on the kindnetd image being embeded into the kind node image so that tag does not really matter since its never requested from ecr. 

When kindnetd "support" for capd clusters was added, the manifest was pulled out of the node image at build time and uploaded to s3 as a yaml file for the cli to use when installing the cni.  A while back, over a year ago, I changed the kind build to build images and binaries separately.  This change had the side effect of breaking the kindnetd manifest because it was only generated during the image build but it was the binary build that actually did the uploading to s3. This resulted in an empty manifest being upload (for other also strange reasons I dont want to go into).

About 6 months ago I changed the kind build [here](https://github.com/aws/eks-anywhere-build-tooling/pull/2356) to use a checked in manifest file with the (good) intention of having us just manage it during version bumps manually.  Well that didnt work... the very first update we did after we forgot to change the manfiest to point to the new 0.20.0 version.

This PRs looks to take us back to a version similar to how we used to do it, but instead of pulling the manifest out of the image, we pull it out of the go file where it is defined as a const.


jaxesn:
This will fix: https://github.com/aws/eks-anywhere/issues/7264

PR #2868, opened 2024-01-31
Add Instructions for Validating Helm Chart

jonathanmeier5:
*Description of changes:*
Add some documentation to reduce friction in the process of developing and testing curated packages helm charts.

Note that this does not test the helm chart's schema, as that would require generating a bundle and installing
in a cluster running the package controller.

However, this does verify that the helm chart itself is valid yaml and install as expected.


jaxesn (commented):
This is a great start! Can you make sure other packages sme's do something similar for theirs?

jonathanmeier5:
> This is a great start! Can you make sure other packages sme's do something similar for theirs?

will do 👍

PR #2870, opened 2024-01-31
Fix metallb patch

abhinavmpandey08:
*Description of changes:*
New version of metallb requires some volumes to be mounted to the speaker pods that the patch was removing.
This PR updates the patch to keep the volume/volumeMounts


jaxesn:
add to the description here and if its appropriate the upgrading section in the readme.

ahreehong (approved):

abhinavmpandey08:
> add to the description here and if its appropriate the upgrading section in the readme.

added the description, nothing to update for readme as we are not bumping the version

PR #2882, opened 2024-02-02
Update upgrader script to upgrade binaries only when there is a version diff 

rahulbabu95:
*Issue #, if available:*
Upgrader pod uses a upgrade.sh script that upgrades all the core kubernetes binaries with the binaries that are already baked into the pod. At present, we upgrade the binaries every time the script is invoked to upgrade a particular component. This change adds logic to the script to check if there is any version diff between the existing binary on the host and the binary baked into the pod before proceeding with the upgrade. 

*Description of changes:*

### DIFF
--- a/projects/aws/upgrader/upgrade.sh
+++ b/projects/aws/upgrader/upgrade.sh
@@ -47,6 +47,7 @@ kubeadm_in_first_cp(){
 
   components_dir=$(upgrade_components_kubernetes_bin_dir)
 
+  if is_version_chage "kubeadm" "version"; then
   backup_and_replace /usr/bin/kubeadm "$components_dir" "$components_dir/kubeadm"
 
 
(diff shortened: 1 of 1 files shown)


jaxesn (commented):
I wouldnt do it this way. Sometimes we rebuild due to new versions of go but that wont result in a new version. Instead we should use the shasum or some other approach. cc @g-gaston

g-gaston:
yeah def use checksums and not versions

Now, do we need this now?
My main concern atomicity. Skipping the upgrade logic for kubeadm if the existing binary matches the new binary is dangerous, the two binaries being the same doesn't guarantee that the upgrade process succeed before. The container can be restarted mid process: after the binary has been copied but before the upgrade logic is run.

IMO the risk here is not worth the optimization benefits. We should write the upgrade logic in a way that is idempotent, so it can be run multiple times. If you want to be able to skip it, we need something smarter than just checking the version/checksum. We need to check the results of the actual upgrade, which involve knowing things about the kubeadm internals, etc. Or we use a "commit" strategy that records the success of each upgrade step for a particular upgrader version.

rahulbabu95:
That's a very valid point. Yeah we need a mechanism to actually verify the upgrade had happened. We were looking to see if we could optimize the upgrade logic but looks like it's more involved than what it appeared to me in the first glance. I will hold this PR for now and discuss on how and if we want to do this at all.

PR #2923, opened 2024-02-14
Go back to capi patch for CRD race condition

g-gaston:
*Description of changes:*
The patch in upstream controller-runtime left one scenario unsolved and we happen to now be hitting it. I'm working on fixing this upstream but in the meantime we are going back to the original patch.

Ongoing upstream fix: https://github.com/kubernetes-sigs/controller-runtime/pull/2687


jonathanmeier5 (approved):
/lgtm
/approve

PR #2924, opened 2024-02-14
Print patched kube-vip manifest to stdout

abhay-krishna:
Print patched kube-vip manifest to stdout so we can verify it in the bootstrap container logs.


abhinavmpandey08 (approved):
/lgtm
/approve
/woof

PR #2929, opened 2024-02-15
Swap capi race condition patch for new CR version

g-gaston:
*Description of changes:*
All the known bugs have been fixed in controller-runtime so now we don't need to detect the error and restart the pod. CR has cut a release yet so capi hasn't bumped its version yet. Bumping our fork before all of that to verify this fix indeed solves the observed race condition.

Latest change to CR that this is including: https://github.com/kubernetes-sigs/controller-runtime/pull/2689


jonathanmeier5 (approved):
/approve
/lgtm

PR #2891, opened 2024-02-05
Handle KubeletCredentialProviders extraargs for in-place upgrades

taneyland:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/2159

*Description of changes:*
`KubeletCredentialProviders` support became GA in k8s v1.26, and the feature gate was removed in k8s v1.28.
When upgrading nodes in-place from 1.25 -> 1.26 -> 1.27 -> 1.28, the 1.28 upgrade will fail with unrecognized feature flag.
Therefore, we should remove this feature gate if it exists on nodes running k8s v1.26 and above.

in 1.25 node: 

```
capv@tneyla-ub-25-hvcvf:~$ cat /etc/sysconfig/kubelet 

KUBELET_EXTRA_ARGS= --feature-gates=KubeletCredentialProviders=true --image-credential-provider-config=/eksa-packages/credential-provider-config.yaml --image-credential-provider-bin-dir=/eksa-packages/
```

After in-place upgrade to 1.26:
```
capv@tneyla-ub-25-hvcvf:~$ cat /etc/sysconfig/kubelet 

KUBELET_EXTRA_ARGS= --image-credential-provider-config=/eksa-packages/credential-provider-config.yaml --image-credential-provider-bin-dir=/eksa-packages/
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2934, opened 2024-02-19
Fix upgrader script to check if kubelet config exists before updating it

abhinavmpandey08:
*Description of changes:*
We added a fix in https://github.com/aws/eks-anywhere-build-tooling/pull/2891 to remove kubelet extra args when upgrading to k8s >= 1.26 as that kubelet extra args is deprecated/removed. The problem is that config file is created by one of the `eksa-packages` pods. So the packages wasn't deployed or failed to deploy, this file wouldn't exist and the in-place upgrade would fail.
This PR adds a check to see if this kubelet config file exists before trying to modify it.

### DIFF
--- a/projects/aws/upgrader/upgrade.sh
+++ b/projects/aws/upgrader/upgrade.sh
@@ -165,13 +165,18 @@ kubelet_and_kubectl() {
 }
 
 update_kubelet_extra_args() {
-  kubelet_extra_args=$(cat /etc/sysconfig/kubelet)
+  kubelet_conf="/etc/sysconfig/kubelet"
+  if [ ! -f kubelet_conf ]; then
+    echo "kubelet config file ${kubelet_conf} not found, skipping"
+    return
+  fi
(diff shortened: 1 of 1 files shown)


projects/aws/upgrader/upgrade.sh
```diff
@@ -165,13 +165,18 @@ kubelet_and_kubectl() {
...
+    return
+  fi
+  kubelet_extra_args=$(cat kubelet_conf)
   feature_gate=" --feature-gates=KubeletCredentialProviders=true"
   if [[ $kubelet_extra_args == *$feature_gate* ]]; then
     kubelet_extra_args=${kubelet_extra_args//"$feature_gate"/}
     mkdir "$components_dir/extraargs"
-    backup_file /etc/sysconfig/kubelet "$components_dir/extraargs"
-    echo "$kubelet_extra_args" > /etc/sysconfig/kubelet
+    backup_file kubelet_conf "$components_dir/extraargs"
```
abhay-krishna, line 178: needs a $

PR #2992, opened 2024-03-08
Attempt to apply patches and generate checksums/attribution if successful

abhay-krishna:
Whenever the project upgrader encounters a project that has patches, it doesn't try to apply them because it could lead to conflicts with the new Git tag. So it just updates the Git tag, Go version and README files for the project and opens a PR with only these changes, which it then marks as an incomplete PR because patches, attribution and checksums need to be updated. 

However, sometimes patches do get cleanly applied across Git tags, especially if they're smaller patches. This PR adds the logic to always apply patches and decide the flow based on the outcome of the patch application. If patches were successfully applied, it proceeds with the generation of attribution and checksums and includes them in the PR, else it does what it does today, i.e., an incomplete PR with a warning, but with additional information about the patch that failed, and the files in the patch that did not apply correctly.

Example: https://github.com/abhay-krishna/eks-anywhere-build-tooling/pull/57

### DIFF
--- a/tools/version-tracker/pkg/commands/upgrade/upgrade.go
+++ b/tools/version-tracker/pkg/commands/upgrade/upgrade.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -31,8 +32,10 @@ import (
 // Run contains the business logic to execute the `upgrade` subcommand.
 func Run(upgradeOptions *types.UpgradeOptions) error {
 	var currentRevision, latestRevision string
-	var projectHasPatches bool
+	var patchApplySucceeded bool
+	var totalPatchCount int
 	var updatedFiles []string
+	patchesWarningComment := constants.PatchesCommentBody
 
 	projectName := upgradeOptions.ProjectName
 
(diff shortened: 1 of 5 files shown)


tools/version-tracker/pkg/commands/upgrade/upgrade.go
```diff
@@ -258,28 +259,34 @@ func Run(upgradeOptions *types.UpgradeOptions) error {
...
-				if !projectHasPatches {
-					if _, err := os.Stat(filepath.Join(projectRootFilepath, constants.ChecksumsFile)); err == nil {
-						logger.Info("Updating project checksums and attribution files")
-						projectChecksumsFileRelativePath := filepath.Join(projectPath, constants.ChecksumsFile)
-						err = updateChecksumsAttributionFiles(projectRootFilepath)
-						if err != nil {
-							return fmt.Errorf("updating project checksums and attribution files: %v", err)
-						}
-						updatedFiles = append(updatedFiles, projectChecksumsFileRelativePath)
+				if projectHasPatches {
```
jaxesn, line 262: If it fails to apply, do you want to try checking in the .rej files? im not sure how helpful they will be but maybe will provide some indication to how big the conflict is?

abhay-krishna: We currently don't generate the .rej files when applying patches, but it's just a matter of adding a flag though. Are you saying we start doing that?

abhay-krishna: The patch will need human intervention anyway so they're going to know when they apply it locally. I don't think it will make a big difference to check it in. Maybe in the patch comment, we can specify how many patches were applied before it failed, and what files had conflicts.

abhay-krishna: Included failed patch number and files information (Example: https://github.com/abhay-krishna/eks-anywhere-build-tooling/pull/57#issuecomment-1989500413) in the warning.

jaxesn (approved):

PR #2988, opened 2024-03-08
Remove patch for building OVAs using content library

abhay-krishna:
This PR removes the patch that configures Packer to upload OVAs to a content library. Since we now build in the Colo lab, we can build them without first exporting to content libraries.

**Testing**

**Built on Colo with content library**
```console
# ls -lh ubuntu-2004-kube-v1.29.1-1.vmdk
-rw-r--r-- 1 1100 1100 1.1G Mar  7 01:38 ubuntu-2004-kube-v1.29.1-1.vmdk
# qemu-img check ubuntu-2004-kube-v1.29.1-1.vmdk
No errors were found on the image.
```

**Built on Colo without content library**
```console
# ls -lh ubuntu-2004-kube-v1.29.1-disk-0.vmdk
-rw-r--r-- 1 1100 1100 1.2G Mar  8 02:06 ubuntu-2004-kube-v1.29.1-disk-0.vmdk
# qemu-img check ubuntu-2004-kube-v1.29.1-disk-0.vmdk
No errors were found on the image.
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2993, opened 2024-03-08
Always delete the Packer builder VM at the end of the build

abhay-krishna:
In our content library patch, we had explicitly set the nested `destroy` field in the content library configuration section to `"true"`, so Packer builder VMs were getting deleted when the build finishes. When we removed the patch in #2988, the build fell back to image-builder's default behavior of leaving around VMs. In this PR, I am patching image-builder so we get the old behavior back.


abhinavmpandey08 (approved):
/lgtm
/woof

PR #3002, opened 2024-03-12
Remove newlines from kubeVersion field

abhay-krishna:
The Kubernetes tag file on GitHub has a trailing newline, so we need to remove it before putting it in the EKS-D latest releases file. Also added logic to update the PR description on an existing PR, in case a new version gets released again when the PR is still open.


pokearu (approved):
/lgtm
/woof

PR #3043, opened 2024-03-30
Revert helm version upgrade

ahreehong:
*Description of changes:*
We removed our patches adding support for the flag `--insecure-skip-tls-verify`, as the functionality had been added to upstream helm. However, upstream has a bug when using proxy with the `--insecure-skip-tls-verify` flag so the helm commands fails. 

Reverting to the previous version of helm. Will investigate the upstream bug further in a separate ticket.


ahreehong:
Created a patch to fix this issue instead of reverting to an older version of helm: https://github.com/aws/eks-anywhere-build-tooling/pull/3046

PR #3071, opened 2024-04-10
Add 1-30 release branch support

rahulbabu95:
*Issue #, if available:*
Add EKS-D 1-30 release branch support for build-tooling projects.

*Description of changes:*


projects/kubernetes-sigs/image-builder/Makefile
```diff
@@ -68,6 +68,7 @@ ADDITIONAL_PAUSE_1-26_FROM=1-26
 ADDITIONAL_PAUSE_1-27_FROM=1-27
 ADDITIONAL_PAUSE_1-28_FROM=1-28
 ADDITIONAL_PAUSE_1-29_FROM=1-28
```
ahreehong, line 70: is there a reason we're still using 1-28 for 1-29 and 1-30?

rahulbabu95: So the pauseVersion between 1-28, 1-29 and 1-30 didn't change from 3.9 in Kubernetes. This can be verified by checking the [value](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.0/cmd/kubeadm/app/constants/constants.go#L436) for different branches 1-28, 1-29 & 1-30. Basically we can use the same value as of 1-28!

ahreehong (approved):

projects/aws/bottlerocket-bootstrap/buildspecs/batch-build.yml
```diff
@@ -47,6 +47,12 @@ batch:
         compute-type: BUILD_GENERAL1_SMALL
         variables:
           RELEASE_BRANCH: 1-29
+    - identifier: aws_bottlerocket_bootstrap_1_30
```
jaxesn, line 50: is adding these a problem if a 1.30 BR release doesnt exist yet?

projects/kubernetes-sigs/image-builder/Makefile
jaxesn: good explanation!

jaxesn (commented):
This seems reasonable to me. Not sure about the BR builds, i feel like we usually have to wait on those?

PR #3072, opened 2024-04-11
Bump Bottlerocket versions to latest release

abhay-krishna:
Bump Bottlerocket versions to the latest upstream release `v1.19.4`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3075, opened 2024-04-11
Fix Bottlerocket host container metadata files

abhay-krishna:
Due to using `yaml` tags in the Go struct, the field names were generated with capital letters in #3072. This PR fixes that issue by changing the struct to use `json` tags. Also fixed a few `go vet` issues.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #3087, opened 2024-04-16
Bump tink to v0.10.0

ahreehong:
*Issue #, if available:*

*Description of changes:*
Bump tink version from `v0.8.0` to `v0.10.0`

Removes `--tls=false` from the tink-server deployment as the flag is removed upstream

We need the patch due to same issue explained here: https://github.com/aws/eks-anywhere-build-tooling/pull/754/files#r865110369. Since we are removing `.ONESHELL`, we need to fix the whitespace for the `out/release/tink.yaml` target


abhay-krishna (approved):
/lgtm
/approve

PR #3109, opened 2024-04-23
Improve build fail remediation logic to factor in newer docker versions

rahulbabu95:
*Issue #, if available:*

The current remediation logic greps for a hardcoded string from the error received from `docker buildx` command. Ideally we want to suggest creating a multiarch driver as a workaround if the buildx uses the default driver. To accomplish this, we could just grep for the key words on the error message instead of the entire string as it might change between different docker versions.

*Description of changes:*
Tested locally by building on the default driver on docker version:  25.0.3


jaxesn (approved):
Nice change!

PR #3191, opened 2024-05-20
Add patch to force-delete Nutanix builder VMs

abhay-krishna:
Nutanix build failures leave behind running VMs on Nutanix Prism central, so this PR adds a patch that force-deletes these VMs even on failure. 


abhinavmpandey08 (approved):
/lgtm
/approve
/woof

PR #3211, opened 2024-05-28
Add retries in case of Nutanix image size mismatch

abhay-krishna:
Nutanix image builds sometimes fail with the flaky issue of image size mismatch. This PR adds a retry for the case when such a failure occurs.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #3234, opened 2024-06-03
Make image-builder upgrade flow non-release-branched

abhay-krishna:
The `kubernetes-sigs/image-builder` project uses the same Git tag for all release branches so we do not need to run the upgrade flow separately for every branch. Hence we're modifying the upgrade batch buildspec to have a single entry for the image-builder upgrade job.


jaxesn (commented):
Is kind the same way?

jaxesn (approved):

PR #3242, opened 2024-06-04
Update kubernetes sigs kind

jaxesn:
*Issue #, if available:*

supersedes: https://github.com/aws/eks-anywhere-build-tooling/pull/3224

I fixed up the patches, the only real change is they added nftables and removed ebtables from the base image. 

I pulled in the new patch from Dims which allows a tar to be passed to the node-image build process.  If you look at the scripts I deleted, we used to have to "fake" the source repo structure because kind only supported building kube from source.  I moved some of that code to main script now so its a bit easier to follow.

In a perfect world, the eks-d server.tgz matches upstream (I am going to look into what this would take) and we could just pass it directly to kind.  Either way, I think this change is at least slightly better because its more clear what is happening.

*Description of changes:*


abhay-krishna (approved):
/approve

PR #3267, opened 2024-06-07
Fix kubeadm upgrade flow for k8s version 1.30 and above

rahulbabu95:
Kubeadm upgrader runs the upgrade apply command by passing in the cluster config using the `--config` flag along with combintaion of `--allow-experimental-upgrades`, `--ignore-preflight-errors=CoreDNSUnsupportedPlugins,CoreDNSMigration`, `--yes`. Kubeadm Introduced a breaking [change](https://github.com/kubernetes/kubernetes/pull/123068) in K8s version 1.30 wherein they introduced a new upgradeConfiguration type and only accepted that as a config for the upgrade apply command. Furthermore, they also deprecated support for multiple other flags and made them configurable via the upgrade config. Now kubeadm upgrade apply command only takes in `kubeadm-config` config map present on the cluster as the source of truth during upgrades and does not accept any local cluster configuration being passed on. This change handles kubeadm upgrades for version 1.30 to transition smoothly from the legacy flow.

*Issue #, if available:*
https://github.com/kubernetes/kubeadm/issues/2823

*Testing:*
Manually tested 1.29 to 1.30 InPlace upgrade on a cluster. 


projects/aws/upgrader/upgrade/kubeadm.go
```diff
@@ -57,35 +65,55 @@ func (u *InPlaceUpgrader) KubeAdmInFirstCP(ctx context.Context) error {
...
 		return execError(kubeAdmVersionCmd, string(version))
 	}
-	logger.Info("current version of kubeadm", "cmd", "kubeadm version", "output", string(version))
+	logger.Info("current version of kubeadm", "cmd", "kubeadm version -oshort", "output", string(version))
 
-	kubeAdmUpgPlanCmd := []string{"kubeadm", "upgrade", "plan", "--ignore-preflight-errors=CoreDNSUnsupportedPlugins,CoreDNSMigration", "--config", newKubeAdmConfig}
-	kubeAdmUpgPlan, err := u.ExecCommand(ctx, kubeAdmUpgPlanCmd[0], kubeAdmUpgPlanCmd[1:]...)
+	// K8s version passed to the upgrader object is of the form vMajor.Minor.Patch-eksd-tag
+	// so it's safe to parse the version
+	kubeVersion, err := strconv.ParseFloat(u.kubernetesVersion[1:5], 64)
```
abhinavmpandey08, line 81: instead of parse float, can we parse the semvar?

projects/aws/upgrader/upgrade/kubeadm.go
```diff
@@ -57,35 +65,55 @@ func (u *InPlaceUpgrader) KubeAdmInFirstCP(ctx context.Context) error {
...
-	kubeAdmVersionCmd := []string{"kubeadm", "version"}
+	kubeAdmVersionCmd := []string{"kubeadm", "version", "-oshort"}
 	version, err := u.ExecCommand(ctx, kubeAdmVersionCmd[0], kubeAdmVersionCmd[1:]...)
 	if err != nil {
 		return execError(kubeAdmVersionCmd, string(version))
 	}
-	logger.Info("current version of kubeadm", "cmd", "kubeadm version", "output", string(version))
+	logger.Info("current version of kubeadm", "cmd", "kubeadm version -oshort", "output", string(version))
 
-	kubeAdmUpgPlanCmd := []string{"kubeadm", "upgrade", "plan", "--ignore-preflight-errors=CoreDNSUnsupportedPlugins,CoreDNSMigration", "--config", newKubeAdmConfig}
```
abhinavmpandey08, line 75: Are we not running upgrade plan anymore?

projects/aws/upgrader/upgrade/kubeadm.go
```diff
@@ -293,3 +377,45 @@ func (u *InPlaceUpgrader) restoreCoreDNSConfig(ctx context.Context, cmpDir strin
 	logger.Info("restored coreDNS config successfully!")
 	return nil
 }
+
+func newTrue() *bool {
+	t := true
+	return &t
+}
+
+func generateKubeAdmConfCM(clusterConfig string) corev1.ConfigMap {
```
abhinavmpandey08, line 386: nit: can you make this func return a pointer
```suggestion
func generateKubeAdmConfCM(clusterConfig string) *corev1.ConfigMap {
```

projects/aws/upgrader/upgrade/kubeadm.go
```diff
@@ -293,3 +377,45 @@ func (u *InPlaceUpgrader) restoreCoreDNSConfig(ctx context.Context, cmpDir strin
...
+func generateKubeAdmUpgradeConfig(version string) UpgradeConfiguration {
+	preflightErrorsList := []string{"CoreDNSUnsupportedPlugins", "CoreDNSMigration"}
+	return UpgradeConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       upgradeConfigurationKind,
+			APIVersion: kubeadmAPIv1beta4,
+		},
+		Apply: UpgradeApplyConfiguration{
+			KubernetesVersion:         version,
+			AllowExperimentalUpgrades: newTrue(),
```
abhinavmpandey08, line 409: nit: can you use `ptr.Bool(true)` instead?
```suggestion
			AllowExperimentalUpgrades: ptr.Bool(true),
```

rahulbabu95: um don't think that's standard lib function. You mean use this: "github.com/aws/eks-anywhere/pkg/utils/ptr" .?

projects/aws/upgrader/upgrade/kubeadm.go
rahulbabu95: i just moved around stuff cos i didn't have to append kubelet config for 1.30 plus versions. so both of these run upgrade plan but at different flow. just some reordering i guess.

projects/aws/upgrader/upgrade/kubeadm.go
abhinavmpandey08: you can use something like this https://go.dev/play/p/PiBOfDuYBRs

abhinavmpandey08 (approved):
/approve
/lgtm

PR #3282, opened 2024-06-11
adds ova validation to check kubelet version

jaxesn:
*Issue #, if available:*

*Description of changes:*

We ran into a case where the image build somehow produced a 1.27 image with 1.25 artifacts in it.  Based on the logs we could find root cause why this happened.  We are thinking it could be something on the vsphere export side.  Adding this should at least tell us during the build if something strange like this happens again.


g-gaston (commented):
/lgtm

PR #3291, opened 2024-06-12
backports latest base image tags

jaxesn:
*Issue #, if available:*

*Description of changes:*
We havent been backporting these and we should be. will talk @abhay-krishna about a better strategy here.


abhay-krishna (approved):
/approve

PR #3319, opened 2024-06-12
fixes release branch check for older kube versions

jaxesn:
*Issue #, if available:*

*Description of changes:*

We added this check for cases where newer branches, like main, supported newer versions of kube than previous release branches. In those cases we want builds for say 1.30 to be skipped on release-0.19 which does not support 1.30

This adds support for checking the supported list so we can not build 1.25 on main, but still build it on release-0.19


abhay-krishna (approved):
/approve

PR #3403, opened 2024-06-26
only search for nongolang licenses in root

jaxesn:
*Issue #, if available:*

*Description of changes:*

The gather-nongolang-licenses was introduced for helm charts, and then later used for the linux build.  The intention is to make sure our produced artifacts, including charts, have the proper upstream license included.  What we were doing makes sense, however, we end up including all the licenses in the `vendor` directory(s).  This we def do not to include all these.  The problem here is the chart ends up getting put into a secret and the recent update to cert manager puts it over the size limit.

This changes the find to only look in the root directory, which should be all we need for the helm repos. I checked each build to ensure we were getting a license file for each helm chart.

### DIFF
--- a/projects/torvalds/linux/Makefile
+++ b/projects/torvalds/linux/Makefile
@@ -3,7 +3,7 @@ GIT_TAG=$(shell cat GIT_TAG)
 
 REPO=linux
 REPO_OWNER=torvalds
-REPO_SPARSE_CHECKOUT=tools lib/bootconfig.c include/linux/bootconfig.h
+REPO_SPARSE_CHECKOUT=COPYING tools lib/bootconfig.c include/linux/bootconfig.h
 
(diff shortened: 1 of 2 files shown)


projects/torvalds/linux/Makefile
```diff
@@ -3,7 +3,7 @@ GIT_TAG=$(shell cat GIT_TAG)
 
 REPO=linux
 REPO_OWNER=torvalds
-REPO_SPARSE_CHECKOUT=tools lib/bootconfig.c include/linux/bootconfig.h
+REPO_SPARSE_CHECKOUT=COPYING tools lib/bootconfig.c include/linux/bootconfig.h
```
jaxesn, line 6: this is the only other project using this license method. while looking at this i noticed since we sparse checkout this repo we werent getting the right license file. this fixes that.

PR #3410, opened 2024-07-08
skip patching KubeVip manifest if its not present in bottlerocket-boo…

banandv:
* skip patching KubeVip manifest if its not present.

For team, like us, who don't use/need KubeVipManifest, `patchKubeVipManifest()` fails because manifest does not exist. This change skips the patching if manifest does not exist without causing failure.

#### Testing
* build the image locally and used it to create an EKS local cluster.


abhay-krishna:
@banandv can you update the checksums file for bottlerocket-bootstrap? You can get the expected checksums from the failed presubmit logs.

PR #3411, opened 2024-07-08
remove uneccessary builds from staging and checksums

jaxesn:
*Issue #, if available:*

*Description of changes:*

While changing some of the presubmits to only run on main, like the packages builds, i noticed there were a couple running in the staging build and a couple builds without go bins running in the checksum builds.


abhay-krishna (approved):
/approve
/lgtm

PR #3472, opened 2024-07-12
[release-0.19] Add lb_class_only env var to kube-vip ds

rahulbabu95:
Kube-vip deployed by tink-stack is trying to manage LoadBalancer IPs of any external service requesting for a LB IP. Ideally kube-vip from tink stack should only manage IPs for those services that have LoadBalancerClass set to `kube-vip.io/kube-vip-class` on the service. This change adds the necessary env in the kube-vip daemon set to only manager the LB Ips for the services with LB class specified to above value. 


ahreehong:
you'll need to bump the chart version to 0.2.5 in Chart.yaml, README.md, and in GIT_TAG to fix the presubmit

jacobweinstock (approved):

PR #3509, opened 2024-07-22
add dhcp project to build tooling

ahreehong:
*Issue #, if available:*

*Description of changes:*
Add dhcp relay project. We need to build this project to support dhcp relay on the tinkerbell provider.


projects/isc-projects/dhcp/Makefile
```diff
@@ -0,0 +1,62 @@
...
+# since this would clobber the other build's artifacts
+UPLOAD_DO_NOT_DELETE=true
+
+
+include $(BASE_DIRECTORY)/Common.mk
+
+
+tarballs: gather-non-golang-licenses $(OUTPUT_DIR)/ATTRIBUTION.txt
+
+$(OUTPUT_BIN_DIR)/%: MAKEFLAGS=
```
jaxesn, line 42: add this, i am doing something similar with ipxe and ran into this on my linux host

# since we are yum installing, we need to run the docker container as root
$(OUTPUT_BIN_DIR)/%: export USER=

jaxesn: actually ignore that for now. i have a better idea, ill update this spot later.

side note: you likely wouldnt see this if you were building on a mac just because of the way users are handled in the builder base docker container.

jaxesn (approved):
Really good work on this!

PR #3530, opened 2024-07-25
adds ipxedust build

jaxesn:
*Issue #, if available:*

*Description of changes:*

This adds the `tinkerbell/ipxedust` project to build tooling and uses the upstream scripts from there to build ipxe, which is a native binary as well as other image formats (iso/img).  Since we dont really have any infra/process to push and pull custom go mod builds, this takes the approach of building ipxe and tar-ing it up as the artifact for the ipxedust build. The actual ipxedust golang code is ignore. 

In the smee build, this artifact is downloaded, using the `PROJECT_DEPENDENCIES` mechanism we use elsewhere. These artifacts are then untared on top of the ipxedust vendor'd directory (boots/vendor/github.com/tinkerbell/ipxedust/binary) so that the binaries that end up in the final build via the go:embed, are the ones we built via eksa build tooling vs the vendored versions from upstream.

The iso and img formats are not used by eksa so they are faked in this build as empty files.  These empty files are what will end up in the final boots/smee bin.


build/lib/prepare_build_container_user.sh
```diff
@@ -31,3 +31,4 @@ groupadd --gid "$GROUP_ID" matchinguser
 useradd  --no-create-home --uid "$USER_ID" --gid "$GROUP_ID" matchinguser
 mkdir -p /home/matchinguser
 chown -R matchinguser:matchinguser /home/matchinguser
+echo "matchinguser ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
```
jaxesn, line 34: this is to make sudo work when running the builder base container on a linux host, we onyl need this when yum installing, which have not done before (except for Ahree's in progress dhcp build)

jacobweinstock (approved):

PR #3585, opened 2024-08-07
Fix tink builds for release-0.19 branch

ahreehong:
*Issue #, if available:*

*Description of changes:*
We update the tinkerbell tink project to build as batch builds with the addition of the dhcp relay related projects. This fixes the builds for the release-0.19 branch. 

We create a batch-build.yml, but only have one job in the batch so that it runs as it had before.


abhay-krishna (approved):
/approve

PR #3607, opened 2024-08-12
Make source images for Nutanix Red Hat builds private

abhay-krishna:
Now that the source images for Nutanix Red Hat builds have been made private, this PR adds logic to source them using a pre-signed S3 URL.


pokearu (approved):
/lgtm

PR #3622, opened 2024-08-13
Add Go version source of truth for tinkerbell hook project

rahulbabu95:
*Issue #, if available:*
`Tinkerbell/hook` project was recently added to upstream projects as part of Tinkerbell stack updates. The upgrader project tracks the go version for a project based on the source of truth it has to look which is supplied through `ProjectGoVersionSourceOfTruth`. This change adds the Go version source of truth and derives it from the hook-bootkit docker [file](https://github.com/tinkerbell/hook/blob/9dc7c7d39742b4d98699848cb57306edb42fd2fa/images/hook-bootkit/Dockerfile#L4).


abhay-krishna (commented):
/approve
/lgtm

PR #3626, opened 2024-08-14
Update schema.json for MetalLB to align with upstream repository

jhaanvi5:
**Description of changes:**
* Update schema.json for MetalLB to align with upstream repository.
* Removed chart corresponding to [frr-k8s-0.0.14.tgz](https://github.com/metallb/metallb/blob/main/charts/metallb/charts/frr-k8s-0.0.14.tgz) in the patch

**Why we are doing this?**
Customer wants to be able to configure interfaces and other attributes which are supported in upstream repository.

Testing

- `helm install -f  values.yaml metallb oci://public.ecr.aws/g0y6o3v0/metallb/metallb --version  0.14.8-3047eee74b2d54f6406a0f469cb3c72fde300822 --set sourceRegistry=public.ecr.aws/g0y6o3v0`

- Tested package installation by generating local bundle with override to registry `public.ecr.aws/g0y6o3v0`.

### DIFF
--- a/projects/metallb/metallb/helm/patches/0001-packages-patches.patch
+++ b/projects/metallb/metallb/helm/patches/0001-packages-patches.patch
@@ -1,4 +1,4 @@
-From cf521735550ac50e7b78d7d3b8dc279642b77f9d Mon Sep 17 00:00:00 2001
+From d7d56fce16c1acc5810ffa7fa65e34a52003a7f8 Mon Sep 17 00:00:00 2001
 From: Jhaanvi Golani <jhaanvi4.golani@gmail.com>
 Date: Fri, 24 May 2024 16:25:03 -0700
 Subject: [PATCH] packages patches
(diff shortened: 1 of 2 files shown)


projects/metallb/metallb/helm/schema.json
```diff
@@ -36,7 +36,7 @@
         "name",
```
vivek-koppuru, line 36: I thought we are using the upstream schema instead of maintaining our own no? Along with setting additionalProperties to true

projects/metallb/metallb/helm/patches/0001-packages-patches.patch
```diff
@@ -93,6 +95,316 @@ index ef8824fe..be8cdd96 100644
  
  
  # A chart can be either an 'application' or a 'library' chart.
+diff --git a/charts/metallb/charts/frr-k8s-0.0.14.tgz b/charts/metallb/charts/frr-k8s-0.0.14.tgz
```
vivek-koppuru, line 98: Do you have steps to document this?

projects/metallb/metallb/helm/schema.json
```diff
@@ -2,188 +2,493 @@
   "$id": "https://packages.eks.amazonaws.com/metallb.schema.json",
```
vivek-koppuru, line 2: Shouldn't this change if we are using upstream schema?

jhaanvi5: It's not exactly same, we do modify it a bit for packages so updated the id.

projects/metallb/metallb/helm/patches/0001-packages-patches.patch
jhaanvi5: Steps are same for creating the patch we are just removing the zip for frr-k8s in chart directory and updating the patch

vivek-koppuru (approved):

PR #3643, opened 2024-08-16
Include branch name other than main in the PR title

sp1999:
*Description of changes:*
The most recent [PR](https://github.com/aws/eks-anywhere-build-tooling/pull/3639) created by `eks-distro-pr-bot` to update ATTRIBUTION.txt files on `release-0.20` branch did not include the branch name in the PR title. We follow the convention to prepend "[BRANCH_NAME] " to the PR title for most of the other automated PRs created by the bot when the PR is being created for any release branch in order to quickly identify the base branch. This PR updates the `create_attribution` function to follow this convention. The `-r` argument is removed for the `pr_title` local variable to allow updating it else it would be a read-only variable and throw an error.

### DIFF
--- a/build/update-attribution-files/create_pr.sh
+++ b/build/update-attribution-files/create_pr.sh
@@ -205,10 +205,14 @@ EOF
 }
 
 function pr::create::attribution() {
-    local -r pr_title="Update ATTRIBUTION.txt files"
+    local pr_title="Update ATTRIBUTION.txt files"
     local commit_message="[PR BOT] Update ATTRIBUTION.txt files"
     local pr_branch="attribution-files-update-$MAIN_BRANCH"
     local -r pr_body=$(pr::create::pr_body "attribution")
+
+    if [ "$MAIN_BRANCH" != "main" ]; then
+        pr_title="[$MAIN_BRANCH] $pr_title"
+    fi
     
     local force_push="true"
     if [ -n "${CODEBUILD_RESOLVED_SOURCE_VERSION:-}" ]; then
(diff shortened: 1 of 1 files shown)


build/update-attribution-files/create_pr.sh
```diff
@@ -205,13 +205,16 @@ EOF
...
 function pr::create::attribution() {
-    local -r pr_title="Update ATTRIBUTION.txt files"
+    local pr_title="Update ATTRIBUTION.txt files"
     local commit_message="[PR BOT] Update ATTRIBUTION.txt files"
     local pr_branch="attribution-files-update-$MAIN_BRANCH"
     local -r pr_body=$(pr::create::pr_body "attribution")
     
     local force_push="true"
     if [ -n "${CODEBUILD_RESOLVED_SOURCE_VERSION:-}" ]; then
+        if [ "${CODEBUILD_RESOLVED_SOURCE_VERSION}" != "$MAIN_BRANCH" ]; then
```
abhay-krishna, line 215: This variable will give you the commit hash in CodeBuild, not the branch. In addition to CodeBuild, this script is also run in the attribution Prowjobs where we use `PULL_BASE_REF` to get the branch name. When invoking the script through CodeBuild too, we pass in the PULL_BASE_REF as an override, so I believe we can just use `PULL_BASE_REF` here. Or the `CODEBUILD_SOURCE_VERSION` like in other places.

We also need to do this check when defining the checksums PR title.

sp1999: You are right. We override the `MAIN_BRANCH` variable with the `CODEBUILD_SOURCE_VERSION` or `PULL_BASE_REF` value if it is non-empty string on lines 36 and 40. So I think I can directly check if $MAIN_BRANCH  != "main" to update the title.

jaxesn (approved):

PR #3654, opened 2024-08-20
Update patch to disable waitForApiext initContainer by default

jhaanvi5:
### Issue:
When installing emissary on airgapped environment, emissary pod was failing with `Init:ImagePullBackOff` error. The root cause of the error was pod spec had initContainer which was trying  `"istio/kubectl:1.5.10"`

### Description of changes:
By default, the initContainer `waitForApiext` is enabled above PR makes it false by default.

### Testing:
Tested the config on existing package with below config, and it skips creating initContainer for `waitForApiext` and installs emissary.
```yaml
apiVersion: packages.eks.amazonaws.com/v1alpha1
kind: Package
metadata:
  name: emissary
  namespace: eksa-packages-persistent-mgmt-ag
spec:
  packageName: emissary
  targetNamespace: emissary-system
  config: |
    waitForApiext:
      enabled: false
```


projects/emissary-ingress/emissary/patches/0001-patching_helm_charts_packages.patch
sp1999, line 1: Is this manually tested and verified that it gets installed in air-gapped env?

projects/emissary-ingress/emissary/patches/0001-patching_helm_charts_packages.patch
```diff
@@ -6886,6 +6886,15 @@ index abe1dcaa2..cf5939e44 100644
    # Pod container image pull policy
    pullPolicy: IfNotPresent
  
+@@ -255,7 +257,7 @@ restartPolicy:
+ terminationGracePeriodSeconds:
+ 
+ waitForApiext:
+-  enabled: true
++  enabled: false
```
vivek-koppuru, line 6894: I understand this is a workaround, but is this the intended behavior we want for the long term?

projects/emissary-ingress/emissary/patches/0001-patching_helm_charts_packages.patch
jhaanvi5: yes this was tested on airgaped cluster, added details above

projects/emissary-ingress/emissary/patches/0001-patching_helm_charts_packages.patch
jhaanvi5: Yeah, the idea is to keep it disabled by default to avoid this failure for users in air-gapped environment, and later we look into building the image and adding it in, if needed.

vivek-koppuru: But do we know what the side effects for removing this is?

PR #3696, opened 2024-08-30
adds linuxkit build

jaxesn:
*Issue #, if available:*

*Description of changes:*

This is the bare minimal to get us started building the various pieces that make up linuxkit and hook's kernel build.  This simply adds the build for the linuxkit bin and the images needed [here](https://github.com/tinkerbell/hook/blob/main/linuxkit-templates/hook.template.yaml). 

I will have a follow that changes the hook build to rely/pull these built images instead of those from upstream.  

I also plan to poke at and gradually update these images to follow our standard process and be al2 based.  Most of these images actually execute on the hook os via containerd as containers.  Those are the ones listed as onboot or services in the template.  For these we should be able to change them to whatever since it does not matter what the host is.  For the images listed in the init section of the template, those are what are added to the "host" itself, so like containerd and runc as well as the base OS, which is currently from upstream.  

/hold


projects/linuxkit/linuxkit/README.md
```diff
@@ -0,0 +1,21 @@
+## **Hook**
```
abhay-krishna, line 1:
```suggestion
## **LinuxKit**
```

abhay-krishna (approved):
/approve

PR #3700, opened 2024-08-30
fixes combine for linuxkit by checking the repo

jaxesn:
*Issue #, if available:*

*Description of changes:*

A quirk about this build is it uses the dockerfiles from upstream.  The few places we use combine have never done this and therefore not needed the repo to be cloned.  Adding this checkout should be fine for those cases and fix the linuxkit build.


abhay-krishna (approved):
/lgtm
/approve

PR #3819, opened 2024-09-30
fixes support for efi on rhel 9 raw builds

jaxesn:
*Issue #, if available:*

*Description of changes:*

We only support efi for rhel 9 and currently if firmware is set to efi when calling image-builder it will end up getting the efi appended again.  This changes the image builder to only support firmware efi for rhel 9 raw.

~We need a change to the codebuild job to set this when building rhel 9 raw.~


2ez4szliu (approved):
lgtm

PR #3834, opened 2024-10-01
boots uses eksa ipxedust

jaxesn:
*Issue #, if available:*

*Description of changes:*

Once we are building ipxedust, this will download it as an artifact from our s3 bucket and use the built ipxe bins instead of the embedded from ipxedust.

This does not currently do anything with the golang src files from ipxedust, those will continue to be pulled via go mods.

Follow up to: #3530 


abhay-krishna (approved):
/approve

PR #3859, opened 2024-10-04
adds source epoch to ipxe build for reproducibility

jaxesn:
*Issue #, if available:*

*Description of changes:*

Set build time based on the ipxedust repo commit timestamp to make builds reproducible.

Going to keep boots set to ignore checksums because updating boots due to a update in ipxe will get a bit problematic.  I started to code out a way for the boots presubmit to build ipxe and then use it so that we can use presubmits to validate checksums like we do for other projects.  That would have made it so whenever ipxe changes the boots pre would fail, cluing us into the need to update the checksum.  The issue is that the boots bin, both the amd and the arm, needs to include the amd and arm versions of ipxe.  We do not have a good way to build both arm and amd in the same prow job so we could build just the amd or just arm, depending on the job, but that means we would have an out of date version of the other which would affect the checksum anyway.

I think this is a fine trade off. We can validate the checksums in ipxe and ignore them in boots.


abhay-krishna (approved):
/approve

PR #3866, opened 2024-10-06
adds builds for golang based linuxkit images

jaxesn:
*Issue #, if available:*

*Description of changes:*

We use linuxkit to build our hook os.  Linuxkit is both a cli and a series of images which are used to make the final hook os image.  We are currently building these linuxkit images based on upstream dockerfiles.  

This PR changes the few that are golang based, sysctl/sysfs/init, to be built using our standard build tooling.  The init image is the main image which makes up the rootfs of the hook image so it still uses the upstream dockerfile but our builds of the golang bin (init/init.rc/service).


projects/linuxkit/linuxkit/patches/0003-pass-cpu-host-when-running-qemu-to-support-testing-a.patch
```diff
@@ -0,0 +1,26 @@
...
+index 7a7066264..bd80762f6 100644
+--- a/src/cmd/linuxkit/run_qemu.go
++++ b/src/cmd/linuxkit/run_qemu.go
+@@ -464,6 +464,9 @@ func buildQemuCmdline(config QemuConfig) (QemuConfig, []string) {
+ 		} else {
+ 			qemuArgs = append(qemuArgs, "-cpu", "cortex-a57")
+ 		}
++	} else {
++		// aws - We require x86_64 to run al23 with its version of glibc
++		qemuArgs = append(qemuArgs, "-cpu", "host")
```
jaxesn, line 20: When testing with a al23 based hook image the default qemu cpu is set to x86_64_v1, but the version of glibc requires v2.  Passing `-cpu host` matches the host version.  This change will not affect our builds only when we run qemu to test it, which is not something our customers do.

projects/tinkerbell/hook/patches/0004-support-overriding-linuxkit-version-to-use-eksa-buil.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/build.sh b/build.sh
+index bebfa11..b45d48c 100755
+--- a/build.sh
++++ b/build.sh
+@@ -29,7 +29,7 @@ declare -g HOOK_LK_CONTAINERS_OCI_BASE="${HOOK_LK_CONTAINERS_OCI_BASE:-"quay.io/
+ declare -g SKOPEO_IMAGE="${SKOPEO_IMAGE:-"quay.io/skopeo/stable:latest"}"
+ 
+ # See https://github.com/linuxkit/linuxkit/releases
+-declare -g -r LINUXKIT_VERSION_DEFAULT="1.5.0" # LinuxKit version to use by default; each flavor can set its own too
++declare -g -r LINUXKIT_VERSION_DEFAULT="${LINUXKIT_VERSION_DEFAULT:-1.5.0}" # LinuxKit version to use by default; each flavor can set its own too
```
jaxesn, line 19: I meant to do this change earlier, this just makes sure that when we run linuxkit we are using the one built from eksa.

jacobweinstock (approved):

PR #3883, opened 2024-10-09
adds builds for opentnpd and dhcpcd for linuxkit images

jaxesn:
*Issue #, if available:*

*Description of changes:*

This adds openntpd and dhcpcd builds to linuxkit.  Neither of these exists in al2/al23 so building from the source. The new base images for these are al2 based.

The openntpd build is not built with `libtls` so the cert validation constraint would not be supported, but based on the ntp config we are using I do not believe this is needed:

```
servers pool.ntp.org
```

I went through the dhcpcd hooks and the additional mtu script linuxkit provides to make sure I included all the utils it needs (cat/rm/set/etc).  This is something we have to pay attention to when updating, I added a note to the hook readme.


projects/linuxkit/linuxkit/dependencies.yaml
```diff
@@ -0,0 +1,6 @@
+DHCPCD:
```
jaxesn, line 1: This is kind of a new pattern, we may want to break these builds out later on, for now we are only going to use them here to produce images we only use during hook build time.  I did model this a bit off how we do the builder-base versions.yaml.  I was thinking this would be easy to add the upgrade script to check for new versions and update.

projects/linuxkit/linuxkit/Makefile
```diff
@@ -91,7 +91,12 @@ $(OUTPUT_BIN_DIR)/linux-%/rngd: EXTRA_GO_LDFLAGS=-extldflags -static
 $(OUTPUT_BIN_DIR)/linux-%/rngd: EXTRA_GOBUILD_FLAGS=-tags netgo,osusergo,static_build
 $(OUTPUT_DIR)/rngd/attribution/go-license.csv: CGO_CREATE_BINARIES=true
 
-ca-certificates/images/% dhcpcd/images/% firmware/images/% getty/images/% modprobe/images/% openntpd/images/%: IMAGE_CONTEXT_DIR=$(REPO)/pkg/$(IMAGE_NAME)
+ca-certificates/images/% firmware/images/% getty/images/% modprobe/images/%: IMAGE_CONTEXT_DIR=$(REPO)/pkg/$(IMAGE_NAME)
+
+dhcpcd/images/% openntpd/images/%: BASE_IMAGE_NAME=eks-distro-minimal-base-glibc
+
+$(call IMAGE_TARGETS_FOR_NAME,openntpd): openntpd-useradd/images/export
+openntpd-useradd/images/export: IMAGE_USERADD_USER_NAME=_ntp
```
abhay-krishna, line 99: Where is this username coming from? Couldn't find it in the upstream Linuxkit openntpd Dockerfile.

projects/linuxkit/linuxkit/Makefile
```diff
@@ -100,6 +105,42 @@ $(OUTPUT_BIN_DIR)/%/usermode-helper: $(GIT_PATCH_TARGET) | $$(call ENABLE_DOCKER
 	LDFLAGS=-static CFLAGS=-Werror $(MAKE) -C $(REPO)/pkg/init usermode-helper; \
 	mv $(REPO)/pkg/init/usermode-helper $@
 
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: MAKEFLAGS=
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: PLATFORM=$(subst -,/,$*)
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: OUTPUT_PATH=$(MAKE_ROOT)/$(OUTPUT_BIN_DIR)/$*/dhcpcd
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: VERSION=$(shell yq '.DHCPCD.version' dependencies.yaml )
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: SHA256SUM=$(shell yq '.DHCPCD.sha256sum' dependencies.yaml )
+$(OUTPUT_BIN_DIR)/%/dhcpcd/usr/sbin/dhcpcd: $(GIT_PATCH_TARGET) | $$(call ENABLE_DOCKER_PLATFORM,$$(PLATFORM))
+	mkdir -p $(OUTPUT_DIR)/dhcpcd; \
```
abhay-krishna, line 114: Maybe we could move this and the one below for openntpd into a common script or Macro

PR #3909, opened 2024-10-15
Handle kubeadm join and etcd join logic when 'ControlPlaneKubeletLocalMode' is enabled 

rahulbabu95:
*Description of changes:*
Kubeadm introduced a [feature-gate](https://github.com/kubernetes/kubernetes/pull/125582) 'ControlPlaneKubeletLocalMode' in 1.31 that allows for specifying local API server for kubelet instead of load balanced control plane endpoint. CAPI defaults this [flag](https://github.com/kubernetes-sigs/cluster-api/pull/10947) to true from K8s version 1.31 and onwards. While bootstrapping BR nodes, we do not [TLSBootstrap](https://github.com/aws/eks-anywhere-build-tooling/blob/0633a73e3961394d30e0812fd8720d87e12d78f6/projects/kubernetes-sigs/cluster-api/patches/0002-Adding-capi-support-for-Bottlerocket.patch#L786) kubelet as a result with this upstream change the join command gets stuck waiting for Kubelet to bootstrap. Skip 'kubelet-wait-bootstrap' phase so that kubeadm join does not error out.    

For stacked etcd, kubeadm doesn't accept config flag for newly added phase 'control-plane-join-etcd', so run join instead and skip all the other phases as a workaround. 

Tested the above change with bundles override and was able to verify that CP nodes were able to join the cluster. For both stacked and unstacked etcd.


sp1999:
We should also update the `control-plane-join/etcd` subphase with the new `etcd-join` phase in the [joinLocalEtcd](https://github.com/aws/eks-anywhere-build-tooling/blob/0633a73e3961394d30e0812fd8720d87e12d78f6/projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_join.go#L132C6-L132C19) function as well as in the [--skip-phases](https://github.com/aws/eks-anywhere-build-tooling/pull/3909) option for the `kubeadm join` command.

Ref:- https://github.com/kubernetes/kubernetes/pull/125582

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_join.go
```diff
@@ -115,9 +121,20 @@ func controlPlaneJoin() error {
...
+	if err != nil {
+		return errors.Wrap(err, "Error comparing kubeadm version with v1.31.0")
+	}
+
 	// finish kubeadm
-	cmd = exec.Command(kubeadmBinary, utils.LogVerbosity, "join", "--skip-phases", "preflight,control-plane-prepare,kubelet-start,control-plane-join/etcd",
-		"--config", kubeadmJoinFile)
+	// 'kubelet-wait-bootstrap' phase was introduced only in 1.31, skip the phase for 1.31 and above
+	if k8s131Compare != -1 {
+		cmd = exec.Command(kubeadmBinary, utils.LogVerbosity, "join", "--skip-phases", "preflight,control-plane-prepare,kubelet-start,control-plane-join/etcd,kubelet-wait-bootstrap",
```
jaxesn, line 132: Instead of the two exec calls, id probably define the string of skipped phases in the if and then just have the exec code once. same for below too

projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_init.go
```diff
@@ -199,3 +194,18 @@ func controlPlaneInit() error {
 	}
 	return nil
 }
+
+func getKubeAdmVersion() (*versionutil.Version, error) {
```
sp1999, line 198:
```suggestion
func getKubeadmVersion() (*versionutil.Version, error) {
```

jaxesn (approved):

PR #3976, opened 2024-10-28
Retry Nutanix image builds in case of internal server errors

abhay-krishna:
This PR adds the logic to retry Nutanix image builds in case of internal server errors, since we have observed that these usually pass when retried.


sp1999:
This PR doesn't need to be backported to release branches since this is a CI script that runs only on main as we build images only on main for testing

PR #4121, opened 2024-12-12
Always download Nutanix source images and delete them after build

abhay-krishna:
Configuring Nutanix builds to always download the source images and delete them after build so that each build doesn't leave behind a copy of it.


projects/aws/image-builder/builder/types.go
```diff
@@ -119,7 +119,6 @@ type NutanixConfig struct {
 	ImageName         string `json:"image_name"`
 	ImageUrl          string `json:"image_url,omitempty"`
 	ImageExport       string `json:"image_export,omitempty"`
-	SourceImageName   string `json:"source_image_name,omitempty"`
```
abhay-krishna, line 122: This was removed because we don't actually use it anywhere in upstream image-builder.

PR #4174, opened 2024-12-26
Add end of standard support dates for all kubernetes versions above 1.28

sp1999:
*Issue #, if available:*
[#2855](https://github.com/aws/eks-anywhere-internal/issues/2855)

*Description of changes:*
This PR adds the end of standard support dates for all Kubernetes versions starting from v1.28 onwards to indicate the start of extended support period for that version. This file will act as a source of truth for the start of extended support period for each supported kubernetes version.

### DIFF
--- a/EKSD_LATEST_RELEASES
+++ b/EKSD_LATEST_RELEASES
@@ -33,16 +33,16 @@ releases:
 - branch: 1-28
   kubeVersion: v1.28.15
   number: 38
-  endOfStandardSupport: 12/31/2024
+  endOfStandardSupport: "2024-12-31"
 - branch: 1-29
   kubeVersion: v1.29.11
   number: 27
-  endOfStandardSupport: 04/30/2025
+  endOfStandardSupport: "2025-04-30"
 - branch: 1-30
   kubeVersion: v1.30.7
   number: 20
-  endOfStandardSupport: 08/31/2025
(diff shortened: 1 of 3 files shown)


EKSD_LATEST_RELEASES
```diff
@@ -33,12 +33,16 @@ releases:
 - branch: 1-28
   kubeVersion: v1.28.15
   number: 38
+  endOfStandardSupport: 12/31/2024
```
abhay-krishna, line 36: Maybe we should use the unambiguous format of `yyyy-mm-dd` here to avoid confusion?

abhay-krishna:
Also the upgrade automation parses and regenerates this file during every new EKS-D upgrade, so to avoid losing the date field, you'll need to add it [here](https://github.com/aws/eks-anywhere-build-tooling/blob/main/tools/version-tracker/pkg/types/types.go#L69) and handle it accordingly.

EKSD_LATEST_RELEASES
sp1999: Agreed, though it needs to be double-quoted in order to parse it as a string when unmarshalling it.
Ref:- https://discourse.getgrav.org/t/international-yaml-date-field/3479/4

abhay-krishna (approved):

PR #4282, opened 2025-01-31
Update hook kernel with missing modules

rahulbabu95:
*Description of changes:*
The recent Hook builds in CI have been creating HookOS kernels that are booting without loading drivers from `modules.dep.bin` properly. The `modules.dep.bin` do not show up on running systems at all. This causes most hardware to fail to load important things like Network Interface Card drivers. 

Troubleshooting pointed us to a potential issue with Yum repo package changes that have caused the HookOS kernel builds to be missing the linkage between the kernel and the `modules.dep.bin`. We don't know definitively what has changed to the packages or why this happened. What we do know through troubleshooting and testing is that adding a few packages resolves the issues. Comparing the logs between previous working kernels and the newer kernels, there is at least one warning related to all this.
```
#20 0.849 Warning: 'make modules_install' requires depmod. Please install it.
#20 0.849 This is probably in the kmod package.
```


jacobweinstock (approved):

jacobweinstock (approved):

rahulbabu95:
The presubmits need to be looked at but they are not related to these changes. Overriding the tests now to unblock main.

jacobweinstock (approved):

PR #4335, opened 2025-02-20
Improve display command for version tracker

abhay-krishna:
Improve display command for version tracker:
1. Versions table shows project versions for all currently-supported EKS-D release branches
2. Versions table includes up-to-date column to denote if project's current version matches latest upstream version
3. Github version retrieval supports projects which don't have a patch version in their semver
4. 


tools/version-tracker/cmd/display.go
```diff
@@ -27,5 +27,4 @@ var displayCmd = &cobra.Command{
 func init() {
```
rajeshvenkata, line 27: Can we add before and after format in description for reference purpose?

PR #4784, opened 2025-08-25
Support vsphere-clone builder

2ez4szliu:
Add support for image-builder CLI to use vsphere-clone builder.

The PR adds make targets for clone build in kubernetes-sigs/image-builder project Makefile to support using vsphere-clone builder, and then add logic in aws/image-builder to use clone build target if using vsphere-clone builder.

We also added one patch to disable firewall service(firewalld on Rhel and ufw on Ubuntu) because if the base template has firewall enabled, it will block certain ports used by kubelet/api-server, or any other services that are deployed to the cluster, this behavior aligns with the iso build produced by image-builder CLI.

The template for base image can be specified in the vsphere json config file:
```
{
  "cluster":"<vsphere cluster used for image building>",
  "convert_to_template":"false",
  "create_snapshot":"<creates a snapshot on base OVA after building if set to true>",
  "datacenter":"<vsphere datacenter used for image building>",
  "datastore":"<datastore used to store template/for image building>",
  "folder":"<folder on vsphere to create temporary vm>",
  "insecure_connection":"true",
  "linked_clone":"false",
  "network":"<vsphere network used for image building>",
  "password":"<vcenter username>",
  "resource_pool":"<resource pool used for image building vm>",
  "username":"<vcenter username>",
  "vcenter_server":"<vcenter fqdn>",
  "vmx_version":"<hardware version of virtual machine>",
  "template": "<template used by clone builder>"
}
```

*Testing:*
Build image-builder cli, and clone a RHEL 9 image with command
`./image-builder build --os redhat --os-version 9 --hypervisor vsphere --release-channel 1-31 --vsphere-config ~/vsphere.json --builder clone`

Created an EKS-A cluster with the clone template.

### DIFF
--- a/projects/kubernetes-sigs/image-builder/Makefile
+++ b/projects/kubernetes-sigs/image-builder/Makefile
@@ -107,7 +107,8 @@ EXT_FROM_FORMAT=$(if $(and $(filter bottlerocket,$2),$(filter-out ova,$1)),img.,
 CONVERT_TO_UPSTREAM_TARGET=$(strip \
 	$(eval _NEW_TARGET:=$1) \
 	$(eval _NEW_TARGET:=$(subst validate-ova-,validate-node-ova-local-,$(_NEW_TARGET))) \
-	$(eval _NEW_TARGET:=$(subst build-ova-,build-node-ova-vsphere-,$(_NEW_TARGET))) \
+	$(eval _NEW_TARGET:=$(subst build-clone-ova-,build-node-ova-vsphere-clone-,$(_NEW_TARGET))) \
+	$(eval _NEW_TARGET:=$(subst build-ova-,build-node-ova-vsphere-,$(_NEW_TARGET)))  \
 	$(eval _NEW_TARGET:=$(subst -redhat-,-rhel-,$(_NEW_TARGET))) \
 	$(eval _NEW_TARGET:=$(subst -cloudstack,-qemu,$(_NEW_TARGET))) \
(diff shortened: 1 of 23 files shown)


projects/kubernetes-sigs/image-builder/Makefile
```diff
@@ -409,13 +429,34 @@ release-image-build-on-metal-%: validate-supported-image-% $(GIT_PATCH_TARGET) $
 
 # These targets all end with <image format>-<os>-<os-version><optional -efi>
 # ex: local-build-ami-ubuntu-2004, s3-artifacts-ova-ubuntu-2004, upload-artifacts-ova-ubuntu-2004, upload-artifacts-ova-ubuntu-2004-efi
+# Special case for clone targets: local-clone-build-ova-redhat-9-efi
+local-clone-build-%: IMAGE_FORMAT=$(word 1,$(subst -, ,$*))
```
abhay-krishna, line 433: Why can't we include this pattern in the below var assignment for the non-clone targets? The values are the same right?

2ez4szliu: I think this is the easiest way even if we are doing some copy pasting work. I have tried using the same pattern and include `clone` in one of the positional args and it makes the logic very complicated. I have also tried to use pattern `local-build-%-clone%`, but then `clone` gets interpreted into os, provider or firmware

abhay-krishna: Makes sense.

PR #4837, opened 2025-09-14
Increase default disk size for raw image-builder builds

snarkychef:
*Description of changes:*

1-33-redhat-8-raw, 1-30-redhat-8-raw, 1-29-redhat-8-raw, 1-28-redhat-8-raw builds are failing due to packer builder vm running out of default configured 6gb disk size, with the latest eks-d release. 

This PR configures the default disk size for raw image building to 6.5gb overriding the [upstream default](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/raw/packer.json.tmpl#L173) of 6gb using [user defined packer config parameter](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/raw/packer.json.tmpl#L16)

### DIFF
--- a/projects/aws/image-builder/builder/constants.go
+++ b/projects/aws/image-builder/builder/constants.go
@@ -10,6 +10,7 @@ const (
 	DefaultAMIVolumeSize           string = "25"
 	DefaultAMIVolumeType           string = "gp3"
 	DefaultAMIManifestOutput       string = "manifest.json"
+	DefaultBaremetalDiskSizeMb     string = "6656"
 
 	// Paths and URLs
 	buildToolingRepoUrl              string = "https://github.com/aws/eks-anywhere-build-tooling.git"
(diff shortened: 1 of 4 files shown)


projects/aws/image-builder/builder/constants.go
```diff
@@ -10,6 +10,9 @@ const (
 	DefaultAMIVolumeSize           string = "25"
 	DefaultAMIVolumeType           string = "gp3"
 	DefaultAMIManifestOutput       string = "manifest.json"
+	DefaultBaremetalDiskSizeMb     string = "6656"
```
2ez4szliu, line 13: I think we don't need to define these 3 default values. We don't want to send some constraint on the min/max size when upstream does not have this limit
Default disk size is set in upstream image-builder in this file https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/raw/packer.json.tmpl#L173

I think what we want to do is to patch this line to use user value or an environment variable , adn set this environment variable in our CI CodeBuild job https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/raw/packer.json.tmpl#L16

PR #5536, opened 2026-06-29
Add nutanix_cluster_uuid to fix image build cluster resolution

peirulu:
## Summary

- Add `nutanix_cluster_uuid` field to `NutanixConfig` struct in image-builder CLI
- Add `nutanix_cluster_uuid` variable to `packer/nutanix/nutanix.json` template
- Add patch to upstream packer template adding `cluster_uuid` builder field

## Problem

All Nutanix image builds are failing with:
```
==> nutanix: Unable to create virtual machine: error waiting for VM creation: 
    task failed: Failed to perform the operation due to an internal error.
```

Root cause: Two clusters in Prism Central share the name `nutanix-ci-cluster`:
- `7887b68c...` — **Prism Central** (cannot run VMs) on `10.70.2.0/24`
- `000628ec...` — **AHV Compute** (4 hypervisor nodes) on `10.70.128.0/24`

Packer resolves `nutanix_cluster_name: "nutanix-ci-cluster"` to Prism Central, which rejects VM creation.

## Fix

The Packer Nutanix plugin supports `cluster_uuid` ([reference](https://github.com/nutanix-cloud-native/packer-plugin-nutanix/blob/main/builder/nutanix/config.go#L126)) which takes precedence over `cluster_name`. This PR adds the plumbing so that when `nutanix_cluster_uuid` is set in the config, it flows through to the Packer template.

### DIFF
--- a/projects/aws/image-builder/builder/types.go
+++ b/projects/aws/image-builder/builder/types.go
@@ -123,6 +123,7 @@ type RhelConfig struct {
 
 type NutanixConfig struct {
 	ClusterName       string `json:"nutanix_cluster_name"`
+	ClusterUUID       string `json:"nutanix_cluster_uuid,omitempty"`
 	ImageName         string `json:"image_name"`
 	ImageSizeGb       string `json:"image_size_gb,omitempty"`
 	ImageUrl          string `json:"image_url,omitempty"`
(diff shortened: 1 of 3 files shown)


peirulu:
/override imagebuilder-1-29-presubmit

since 1-29 will be deprecated after v0.26.0 is out

2ez4szliu (approved):

PR #5578, opened 2026-07-10
fix(tinkerbell/charts): Increase default memory limit from 128Mi to 256Mi

peirulu:
## Summary
- Add patch to increase default memory limit from 128Mi to 256Mi
- Targets the unified chart at `projects/tinkerbell/charts`
## Why
Tinkerbell pods OOMKill (exit code 137) during EKS Anywhere upgrades because baseline memory usage (~136Mi) exceeds the 128Mi limit. Customer manually increased to 256Mi and confirmed it resolves the issue.

Ticket: https://t.corp.amazon.com/V2262142014

## Test plan
- [x] Ran `make helm/build` — chart builds successfully with patch applied
- [x] Verified built chart output confirms 256Mi memory limits in all sub-charts:
  ```
  rufio/values.yaml:    memory: 256Mi
  smee/values.yaml:     memory: 256Mi
  hegel/values.yaml:    memory: 256Mi
  tink/values.yaml:     memory: 256Mi (controller)
  tink/values.yaml:     memory: 256Mi (server)
  ```


projects/tinkerbell/tinkerbell-chart/chart/values.yaml
rahulbabu95, line 1: I believe this is the legacy chart. we have problem with the new unified repo? the new charts are here: https://github.com/aws/eks-anywhere-build-tooling/tree/main/projects/tinkerbell/charts/helm/patches#:~:text=helm-,patches,-0001%2DLoad%2Dstream? also good to check if it can be done during the runtime: https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/stack/stack.go ?

peirulu: the `createValuesOverride` function only overrides images, networking, tolerations, affinities, and feature flags.

projects/tinkerbell/tinkerbell/helm/patches/0002-Increase-default-memory-limit-from-128Mi-to-256Mi.patch
```diff
@@ -0,0 +1,35 @@
...
+@@ -156,10 +156,10 @@ deployment:
+   resources:
+     limits:
+       cpu: 500m
+-      memory: 128Mi
++      memory: 256Mi
+     requests:
+       cpu: 500m
+-      memory: 128Mi
++      memory: 256Mi
```
rahulbabu95: let's keep the requests as is and just bump the limits?

peirulu: sure

PR #5621, opened 2026-08-07
[release-0.26] Update CHECKSUMS for new builder-base

rahulbabu95:
Regenerates binary checksums against builder-base `standard-ec39a22a692a667411d582405edec7f191486e52.2`.

## Why

The new builder-base ships a different Go toolchain, so compiled binaries differ from the checksums committed against the previous base image (`standard-3a28c2542ca6192ddba95caa53baed9d7a4c8a0d.2`). Every `validate-checksums` step therefore fails, which blocks the staging bundle build and the staging CLI release.


rahulbabu95:
Overriding cluster-api-provider-nutanix-tooling-presubmit. It's failing because the latest builder base doesn't ship with 1.23 and one of the pkg dependcy in 1.25 isn't backward compatible with version 1.23!

PR #5624, opened 2026-08-07
[release-0.26] Update CHECKSUMS for containerd, bottlerocket-bootstrap, packages, hook, upgrader

rahulbabu95:
Second round of checksum updates for builder-base `standard-ec39a22a692a667411d582405edec7f191486e52.2`, following #5621.

## Why a second round

The projects here were **STOPPED** (cancelled) in the earlier staging batches, so they never emitted recomputed checksums and were not covered by #5621.

PR #506, opened 2022-03-03
Cert manager ctl

TerryHowe:
Add cert-manager-ctl to helm chart

PR #5651, opened 2026-09-03
Forward-port CHECKSUMS from release-0.26 to main

rahulbabu95:
## Problem

`validate-checksums` fails on `main` for any PR that triggers those presubmits — 16 failing jobs on #5639 (the EKS Distro release bump for v0.26.2), and the 5 `cloud-provider-aws` jobs have been red on #5244 daily for a month.

Root cause: the builder-base checksum fixes landed on `release-0.26` only (#5621, #5624) and were never forward-ported to `main`.

## What this does

56 CHECKSUMS files, from two sources.

**51 forward-ported from `release-0.26`.** Restricted to projects whose *only* difference between `main` and `release-0.26` is the CHECKSUMS file itself. Verified against the checksums computed by the failing presubmits on #5639 — all match. Example, `cloud-provider-vsphere/1-30`:

```
computed by presubmit  48a0f85e4021d42b… / 75a9a75f88102b94…   == release-0.26
main (stale)           f8ff21d67f32dda9… / a3f3ef81b79d0baf…
```

**5 `cloud-provider-aws/1-30`…`1-34` set to presubmit-computed values.** These were stale on *both* branches, so there was no correct value to port. Values harvested from the `cloud-provider-aws-1-3X-tooling-presubmit` runs; `GIT_TAG` in each build log matches `main`, and for `1-30`/`1-31`/`1-32` two independent runs (#5244 and #5651) computed byte-identical hashes, which is what makes the `1-33`/`1-34` values from the same run trustworthy.

No artifact paths change — every modified line is a hash swap.

## Deliberately excluded

11 files whose projects have other build-input drift (`GIT_TAG`, `GOLANG_VERSION`, `patches/` or `Makefile`), so ported values would not reproduce and a real rebuild is needed:

`autoscaler/1-35`, `autoscaler/1-36`, `aquasecurity/trivy`, `goharbor/distribution`, `goharbor/harbor-scanner-trivy`, `metallb/metallb`, `cluster-api-provider-nutanix`, `prometheus/node_exporter` (covered by #5598), `tinkerbell/hook`, `tinkerbell/rufio`, `tinkerbell/tinkerbell`

Plus **`prometheus/prometheus`**: its presubmit uniquely runs the AL2023 builder-base (`standard-ec39a22a….2023`, from aws/eks-anywhere-prow-jobs#505) while every other job runs `.2`, so AL2023-built binaries hash differently and no ported value can satisfy it. Needs #5630 (the `CHECKSUMS_BUILDSPEC_IMAGE` override) first, then a regenerate.

## Note on ordering

These values are correct for the live builder-base `standard-ec39a22a…`. aws/eks-anywhere-prow-jobs#506 moves builder-base to `standard-930cea64…`, which is a golang bump (1.25.12 to 1.25.13, 1.26.5 to 1.26.7, plus 1.27.0) and will invalidate all of these again. If #506 merges first it should carry its own regenerated checksums.

## Known-unrelated failures on this PR

- arm64 jobs (`containerd-*-arm64`, `linuxkit-arm64`, `runc-arm64`, `dhcp-arm64`, `linux-bootconfig-arm64`) fail deterministically on every PR — the mirrored Prow pod-utility images are amd64-only.
- The first run was largely wiped by an AWS STS `503 ServiceUnavailable` storm; 33 jobs died in `initupload` on `AssumeRoleWithWebIdentity` before the build container started, producing no build log.


rahulbabu95:
/override runc-tooling-presubmit-arm64 linuxkit-tooling-presubmit-arm64 linux-bootconfig-presubmit-arm64 dhcp-presubmit-arm64 containerd-tooling-1-35-presubmit-arm64 containerd-tooling-1-34-presubmit-arm64 containerd-tooling-1-33-presubmit-arm64 containerd-tooling-1-32-presubmit-arm64 containerd-tooling-1-31-presubmit-arm64 containerd-tooling-1-30-presubmit-arm64

rahulbabu95:
/override emissary-tooling-presubmit harbor-tooling-presubmit

